PR #69, opened 2018-08-21
frontend boilerplate with eslint, vuex, vue-router, LESS

NunoDasNeves:


DarkPurple141:
Another thing is you maybe should've let them add the test suites in.

frontend/tests/e2e/specs/test.js
```diff
@@ -0,0 +1,8 @@
+// https://docs.cypress.io/api/introduction/api.html
+
```
DarkPurple141: Test now fails XD. But dw.

PR #85, opened 2018-08-21
Streamlined Backend/Frontend, Testing

DarkPurple141:
Lots of changes, but mostly just deleting things and moving things to different folders.


PapaPwn:
This seems like a lot of versioning and configuration stuff that I don't quite understand. Is there any particular area that needs closer inspection for review?

PR #94, opened 2018-08-23
for luke

DarkPurple141:
Minor changes for @PapaPwn and @Aruzal


DarkPurple141:
If this problem persists, maybe add pack-lock.json to .gitignore for a bit

PR #95, opened 2018-08-25
Partly dealing with Feed requirements

DarkPurple141:
My feeling is this can be merged before we continue on building things out further. Nunez plenty to review and comment or question!


DarkPurple141:
@NunoDasNeves just add comments whereever things feel whack and I'll try to explain. Better now than later. Also easier within context of PR.

frontend/src/router.js
```diff
@@ -18,6 +19,14 @@ export default new Router({
       // this generates a separate chunk (about.[hash].js) for this route
       // which is lazy-loaded when the route is visited.
       component: () => import(/* webpackChunkName: "about" */ './views/About.vue')
+    },
+    {
+      path: '/question/:id',
+      name: 'question',
+      props: ({params: { id }}) => ({
```
NunoDasNeves: So we need to pass the id as a prop to this route, right? What does this function do and why?

PR #96, opened 2018-08-25
Sample API Design

Aruzal:


docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+    "meta": {
+        "user_id": <key>
+        "course_id": <key>
+        "upvotes" <number>
+        "num_answers": <number> // Can be used to determine next answer id
+    }
+    "title": <string>
+    "body": <string>
+}
+```
```
DarkPurple141: questions should have answers as children.

docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+### Universities
+
+Get data for a specific university
+* /api/uni/:id
+
+Get ALL courses for a university
+* /api/uni/:uid/courses
+
+Get (n) courses for a university
+* /api/uni/:uid/courses?n=10
```
PapaPwn: might be better for us to pagination on the backend, so p=1, p=2... is page 1 of course, page 2 of courses, etc.

docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+
+Get (n) questions for a course
+* /api/course/:cid/questions?n=10
+* Top (n) questions ranked by reputation?
+
+Get ALL reviews for a course
+* /api/course/:cid/reviews
+
+Get (n) questions for a course
+* /api/course/:cid/reviews?n=10
```
PapaPwn: same comment as above.

docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+### Questions
+
+Get the question data for a specific question id
+* /api/course/:cid/question/:qid
+
+Get ALL answers for a question
+* /api/course/:cid/question/:qid/answers
+
+Get (n) answers for a question
+* /api/course/:cid/questions/:qid/answers?n=10
```
PapaPwn: ^^^

docs/api.md
DarkPurple141: Agree.

docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+* Answers (Replies to Questions)
+* Reviews
+* Replies (Replies to Reviews)
+
+## Sample Schema's (json)
+
+### User
+```
+"user": {
+    "uid": <unique int>
```
DarkPurple141: all ids inside objects are implicity the type of id, no need to make them specifc. Will just make frontend logic more complex.

DarkPurple141: eg. User
```json
{
   "id" : "xxxx",
   "username" : "xxxx",
   "meta" : { }
}
```

docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+    "profile": {
+        // Stuff for any profile data ect..
+    }
+}
+```
+
+### Universities
+```
+"university" {
+    university_id: <unqiue int>
```
DarkPurple141: same as above with user.

docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+    "name": <string>
+    "img_url" <string>
+}
+```
+
+### Courses
+```
+"course": {
+    "course_id": <unqiue int>
```
DarkPurple141: Same as above with user

docs/api.md
DarkPurple141: I should add on this, from the database perspective answers should just store a single parent id, and then you do a join, but the json response should definitely at least include the ids of child answers.

docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+        "review_id": <key>
+        "upvotes" <number>
+    }
+    "title": <string>
+    "body": <string>
+}
+```
+## Sample API Routes
+
+### Users
```
DarkPurple141: This may have to be extended later.

docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+Get the data for a specific user
+* /api/user/:id
+
+### Universities
+
+Get data for a specific university
```
DarkPurple141:
```
/api/uni should provide all universities
```

DarkPurple141: (paginated)

docs/api.md
DarkPurple141: +1

DarkPurple141: unspecified all.

docs/api.md
```diff
@@ -0,0 +1,239 @@
...
+        "course_id": <key>
+        "upvotes" <number>
+        "num_replies": <number> // Can be used to determine next reply id
+    }
+    "title": <string>
+    "body": <string>
+}
+```
+
+### Replies
```
DarkPurple141: Wonder whether having replies as their own schema is useful. Might just add complexity. Is there anything any way a reply differs from an answer?

DarkPurple141:
Feel free to merge

PR #97, opened 2018-08-26
S2 frontend single question view

NunoDasNeves:
Not actually done yet... Will add input for submitting answers


frontend/src/components/AnswerCard.vue
```diff
@@ -0,0 +1,50 @@
...
+}
+</script>
+
+<style scoped>
+.date {
+  font-size: 0.75em;
+  color: rgba(0, 0, 0, 0.65);
+}
+
+.card {
```
DarkPurple141: likely to abstract part of this out in next PR.

frontend/src/views/Question.vue
```diff
@@ -1,17 +1,24 @@
...
+      <ul>
+          <li v-for="answer in answers" :key="answer.id">
+              <AnswerCard :answer="answer"/>
+          </li>
+      </ul>
     </div>
 </template>
 
 <script>
+import QuestionCard from '@/components/QuestionCard.vue'
```
DarkPurple141: vue file extension not required in imports. (Webpack is able to surmise)

frontend/src/views/Question.vue
```diff
@@ -26,3 +33,9 @@ export default {
   }
 };
 </script>
+
+<style scoped>
```
DarkPurple141: probably should put normalised style like this into root.

DarkPurple141:
Just added some stuff to this branch Nunez.

NunoDasNeves:
I still need to write tests I suppose ... will tackle tomorrow

frontend/src/views/Question.vue
```diff
@@ -33,6 +41,37 @@ export default {
...
+  },
+  methods: {
+    submitAnswer : function (answerForm) {
+        // check that they actually typed something
+        if (answerForm.body == "") {
+            this.answerFormResponse.text = "Please type an answer!"
+            this.answerFormResponse.style = {'form-success': false, 'form-failure': true}
+            return
+        }
+        // post the answer to the api
```
DarkPurple141: This should really be handled in the store.

DarkPurple141: Also strictly beyond scope of current issue.

frontend/src/views/Question.vue
```diff
@@ -33,6 +41,37 @@ export default {
...
+  data: function () {
+    return {
+      answerFormResponse: {
+        text: "",
+        style: {'form-success': false, 'form-failure': false}
+      }
+    }
+  },
+  methods: {
+    submitAnswer : function (answerForm) {
```
DarkPurple141: this syntax is a bit old worldy. New syntax for that kind of func is just:

```js
submitAnswer(args) {

}
```

frontend/src/router.js
```diff
@@ -23,10 +23,8 @@ export default new Router({
     {
       path: '/question/:id',
       name: 'question',
-      props: ({params: { id }}) => ({
-        id
-      }),
-      component: () => import(/* webpackChunkName: "about" */ './views/Question.vue')
+      props: ({params: { id }}) => { id },
```
DarkPurple141: This is a bug as it stands.
() => {} // doesn't return an object
() => ({}) returns an empty object

frontend/src/utils/api.js
```diff
@@ -35,6 +36,15 @@ export function postQuestion (body) {
 }
 
 /**
+ * @param String id the id of the question being answered
```
DarkPurple141: JSDoc format is @param {type} name description

frontend/src/components/AnswerForm.vue
```diff
@@ -0,0 +1,42 @@
...
+
+<script>
+import Card from './Card'
+
+export default {
+  name: 'AnswerForm',
+  components: {
+      Card
+  },
+  data: function () {
```
DarkPurple141:
```js
data() {
  return {
    ...
   }
}
```

frontend/src/router.js
DarkPurple141: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions

NunoDasNeves:
hnnnnnnnng

frontend/src/router.js
NunoDasNeves: oic

PR #102, opened 2018-08-29
S2 backend rcm

PapaPwn:
Refactored routes, controllers and models - RCM


DarkPurple141:
Might wanna resolve those conflicts first fam.

backend/src/controllers/course.js
```diff
@@ -1,36 +1,24 @@
...
+const qdb = require('../models/db-questions.js')
+
+/* Get specifc course data */
+exports.getCourse = function(req, res) {
+    const cid = req.params.cid
+    res.send(`<h1>Course Data for Course #${cid} </h1>`)
+}
+
+/* Get all questions for a course */
+exports.getCourseQuestions = function(req, res) {
```
DarkPurple141: semis?

DarkPurple141: also given you don't use the req directly at all you can do this;

```js
exports.getCourseQuestions = function({ params, query }, res) {
```

backend/src/controllers/question.js
```diff
@@ -1,21 +1,18 @@
...
-    //let qid = req.params.id;
-    //let page_id = req.query.p;
-
-    res.send('<h1>Question Pages</h1>')
-})
-
-module.exports = router
\ No newline at end of file
+/* GET question data. */
+exports.getQuestion = function(req, res) {
```
DarkPurple141: Same comment as above, don't use req directly.

backend/src/models/db-questions.js
```diff
@@ -1,79 +1,43 @@
 const sqlite3 = require('sqlite3')
 const db = new sqlite3.Database('./db/questions.db')
 
-function getAllQuestions() {
-  // Lookup all questions
+function getQuestions(course_id, page_id) {
```
DarkPurple141: camelCase

backend/src/models/db-questions.js
```diff
@@ -1,79 +1,43 @@
...
+  // TODO - PAGING (log avoids unused variable)
+  console.log(page_id)
```
DarkPurple141: if you have this issue you can do
// eslint-disable-line

backend/src/controllers/question.js
```diff
@@ -1,21 +1,18 @@
...
+exports.getQuestion = function(req, res) { 
+    const course_id = req.params.cid
+    const question_id = req.params.id
+    qdb.getQuestion(course_id, question_id)
+        .then(data => res.json(data))
+}
+  
+/* GET question ansewrs. */
+exports.getQuestionAnswers = function(req, res) { 
+    const course_id = req.params.cid
```
DarkPurple141: camelCase

backend/src/models/db-questions.js
```diff
@@ -1,79 +1,43 @@
...
-        })))
-      }
+      'SELECT * FROM answers WHERE cid=? AND qid=?',
+      [course_id, question_id],
+      (err, rows) => { err ? reject() : resolve(rows) }
     )
   })
 }
 
 module.exports = {
```
DarkPurple141: is there a reason why we're doing module.exports = {} for some and exports.func for others? XD

backend/src/models/db-questions.js
```diff
@@ -1,76 +1,45 @@
...
-                })))
-            }
+            'SELECT * FROM answers WHERE cid=? AND qid=?',
+            [course_id, question_id],
+            (err, rows) => { err ? reject() : resolve(rows) }
         )
     })
 }
 
 module.exports = {
```
DarkPurple141: Should be consistent with exports. As per controllers think exports.func makes more sense here.

backend/src/models/db-questions.js
```diff
@@ -1,76 +1,45 @@
...
             }
         )
     })
 }
 
-function getAnswers (questionID) {
-    // Lookup answers
+function getAnswers (course_id, question_id, page_id) {
+    // TODO - PAGING (log avoids unused variable)
+    console.log(page_id)
```
DarkPurple141: can always use // eslint-disable-line
for these ones.

backend/src/controllers/question.js
```diff
@@ -1,19 +1,18 @@
-const express = require('express')
-const router = express.Router({ mergeParams: true })
+const qdb = require('../models/db-questions.js')
 
-/* Get the question data for a specific question id */
-router.get('/:id', function (req, res) {
-    // let qid = req.params.id;
+/* GET question data. */
+exports.getQuestion = function (req, res) {
```
DarkPurple141: given we don't use the req object directly you could rewrite like this anytime this occurs.

```js
exports.getQuestion = function ({params}, res) { // <-- this is 'destructuring'
    // do stuff with param
}
```

PR #116, opened 2018-08-30
S3 dev stable

PapaPwn:
Not quite stable yet, but very close. Just need to test api endpoints with the new data and get questions+answers rendered on the front end.


backend/src/controllers/course.js
```diff
@@ -1,24 +1,27 @@
...
 /* Get all questions for a course */
-exports.getCourseQuestions = function (req, res) {
-    const cid = req.params.cid
-    const pid = req.query.pid
+exports.getCourseQuestions = function ({ params, query }, res) {
+    const courseID = params.id
+    const pageNumber = query.p
 
-    qdb.getQuestions(cid, pid)
+    questionModel.getQuestions(courseID, pageNumber)
```
DarkPurple141: I'd forgotten that if you pass a function signature to a then it will implicitly execute the function with the data you pas it. So. 

Any time you aren't doign a strict transform of the data in the then part and are just sending the json you can do this.

```js
doPromiseThing()
  .then(res.json)
```

DarkPurple141: or even crazier. You could unpack json from the res object in the params and then do just.

```js
function({params, query}, {json}) {
    qdb.getQuestions()
       .then(json)
}
```

backend/src/models/answer.js
```diff
@@ -0,0 +1,17 @@
+const db = require('./db')
+
+function getAnswers (questionID, pageNumber) {
```
DarkPurple141: module.exports = getAnswers

backend/src/models/answer.js
```diff
@@ -0,0 +1,17 @@
...
+    return new Promise((resolve, reject) => {
+        db.all(
+            'SELECT * FROM answer WHERE questionID=?',
+            [questionID],
+            (err, rows) => { err ? reject(err) : resolve(rows) }
+        )
+    })
+}
+
+module.exports = {
```
DarkPurple141: I guess it's a question of consistency. If these types of functions are normal exports just do exports...

backend/src/models/db.js
```diff
@@ -1,125 +1,229 @@
...
+    const review = {
+        body: 'Would bang out of 10!'
+    }
+
+    const reply = {
+        body: 'Good to know... *** sick bastard ***'
+    }
+
+    /* Insert the fake data in to the database */
+    const promises = []
```
DarkPurple141: This seems quite odd. Is this because some updates depend on the next? If the updates are required to be synchronous chain the promises. `Promise.all()` is if the updates both need to be done for some data transform. Also simplifies error handling; you can just follow with a single catch.

backend/src/models/db.js
```diff
@@ -1,125 +1,229 @@
...
-    // Create the courses table
-    db.run(`CREATE TABLE courses (
-        cid INTEGER,
-        code TEXT,
-        name TEXT,
-        faculty TEXT,
-        rating INTEGER,
-        uni INTEGER
+
+const databaseName = process.env === 'test' ? ':memory:' : ':memory:'
```
DarkPurple141: We can't do this in production as if the server crashes we lose our entire database. We can do this in dev. Maybe better to add a switch statement.

DarkPurple141: Currently obviously not a problem.

backend/src/models/db.js
```diff
@@ -1,125 +1,229 @@
...
-    })
-})
+    // Reply inserted
+    promises.push(Promise.all(promises)
+        .then(function ([userID, uniID, courseID, questionID, answerID, reviewID]) {
+            reply.reviewID = reviewID
+            reply.userID = userID
+            return exports.insertDB(db, 'reply', reply)
+        })
+        .catch((err) => console.log(err))
```
DarkPurple141: .catch(console.warn) is suffice for now.

DarkPurple141: But again this should be done differently.

backend/src/models/db.js
```diff
@@ -1,125 +1,229 @@
...
-    (err) => {
-        if (err) {
-            return console.log(err.message)
-        }
-        console.log('Successfully added test data')
-    }
+
+    // Review inserted
+    promises.push(Promise.all(promises)
+        .then(function ([userID, uniID, courseID, questionID, answerID]) {
```
DarkPurple141: Some are arrow functions some aren't :-/

backend/src/routes/question.js
```diff
@@ -1,5 +1,5 @@
 const express = require('express')
-const router = express.Router({ mergeParams: true })
```
DarkPurple141: no longed needed?

backend/src/models/db.js
PapaPwn: Yeah so things like a 'question' depends on there being a 'user' and a 'course', then an 'answer' depends on there being a 'user' and a 'question'. Just did some hacky system to make it synchronous and access all the past promises. Wasn't sure how to do it cleanly because you can only get the previous promises', result when chaining (not the ones before that)?

backend/tests/routes/questions.js
```diff
@@ -0,0 +1,12 @@
+const app = require('../../src')
+const supertest = require('supertest')(app)
+
+/* TODO build out tests to also include catching expected JSON response */
+describe('Question route testing', function () {
+    it('/api/question/1', function (done) {
+        supertest
+            .get('/api/question/1')
```
NunoDasNeves: i seeee kinda. yeah

backend/src/routes/question.js
Aruzal: Yeah. Originally we had a compound key for questions and it was kind of hacky. Since everything now has a unqiue ID you don't necessarily need the course ID to access question data. The course ID can still be sent along with POST's ect.

backend/src/models/db.js
```diff
@@ -1,125 +1,229 @@
...
-    (err) => {
-        if (err) {
-            return console.log(err.message)
-        }
-        console.log('Successfully added course data')
-    }
+function createQuestionTable (db) {
+    db.run(`CREATE TABLE question (
+        questionID INTEGER PRIMARY KEY AUTOINCREMENT,
+        courseID INTEGER NOT NULL,
```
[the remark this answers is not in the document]
Aruzal: What would the purpose of a university ID be used for? It may be useful for some kind of analysis but questions are kind of dependant on the course?

backend/src/utils/helpers.js
```diff
@@ -0,0 +1,5 @@
+
+exports.setData = (responseObject) =>
```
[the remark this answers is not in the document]
Aruzal: Also curious @DarkPurple141

backend/src/models/db.js
```diff
@@ -1,125 +1,229 @@
...
-        qid INTEGER,
-        cid INTEGER,
-        uid INTEGER,
-        likes INTEGER,
-        title TEXT,
-        body TEXT)`
+function createAnswerTable (db) {
+    db.run(`CREATE TABLE answer (
+        answerID INTEGER PRIMARY KEY AUTOINCREMENT,
+        questionID INTEGER NOT NULL,
```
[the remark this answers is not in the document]
PapaPwn: I think the idea is you can backtrace it with question -> course -> uni...

backend/src/models/db.js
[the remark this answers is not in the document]
PapaPwn: I think the idea is you can backtrace to the uni via the course table.

backend/src/routes/question.js
PapaPwn: Yeah cause we now get questions at /api/question rather than api/course/:id/question, with question ids being guaranteed to be unique across all courses.

backend/src/controllers/course.js
PapaPwn: Hmmm, couldn't get this to work.

backend/src/utils/helpers.js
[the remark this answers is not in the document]
PapaPwn: for testing @DarkPurple141 ?

backend/package.json
```diff
@@ -4,9 +4,9 @@
   "private": true,
   "scripts": {
     "start": "npx nodemon ./src/bin/www",
-    "lint": "npx eslint src/**/*.js tests/**/*.js",
+    "lint": "npx eslint src/**/*.js tests/**/*.js --fix",
     "test:src": "npx nyc mocha tests/**/*.js",
-    "test": "npm run lint && npm run test:src",
+    "test": "export NODE_ENV=test npm run lint && npm run test:src",
```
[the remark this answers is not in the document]
DarkPurple141: yes.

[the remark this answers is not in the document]
DarkPurple141: or alternatively app.get('env') => NODE_ENV

backend/src/controllers/course.js
DarkPurple141: It should :-/

DarkPurple141:
@PapaPwn @Aruzal, made a couple more changes

PR #118, opened 2018-09-01
Rework to allow for dynamic actions in Vuex store

DarkPurple141:
@NunoDasNeves This is what we were after.


backend/src/models/db.js
```diff
@@ -173,14 +173,16 @@ function devInitDB (db) {
             // uni dependencies
             comp4920.universityID = uniID
 
-            ;[question, answer, review, reply].reduce((_, curr) => { curr.userID = userID })
+            ;[question, answer, review, reply]
+                .reduce((_, curr) => { curr.userID = userID }, question)
```
NunoDasNeves: what's this doing?

frontend/src/store/questions/constants.js
```diff
@@ -0,0 +1,23 @@
+import {
+  getQuestions,
+  getQuestion,
+  getAnswers
+} from '@/utils/api/questions'
+
+export const ACTIONS = {
+  GET_FEED: Symbol('getQuestionFeed'),
```
NunoDasNeves: idk what any of this is...
Also I think get_feed isn't descriptive enough; get_questions or get_question_feed is better

frontend/src/store/questions/index.js
```diff
@@ -16,17 +20,9 @@ const state = {
...
-      questionID,
-      title,
-      body,
-      author: userID,
-      published: format(Date.now(), 'DD/MM/YY')
-    }))
-  },
-  question: ({questionObj: {question}}) => question,
-  answers: ({questionObj: {answers}}) => answers,
+  feed: ({questions}) => questions.map(questionMapper),
```
NunoDasNeves: Again, I think 'feed' isn't descriptive enough, because it could refer to many things. 'questionFeed' or 'questions' is better imo

frontend/src/store/questions/constants.js
```diff
@@ -0,0 +1,23 @@
...
+  GET_QUESTION: Symbol('getQuestion'),
+  GET_ANSWERS: Symbol('getAnswers')
+}
+
+export const REQUEST = {
+  [ACTIONS.GET_FEED]: getQuestions,
```
NunoDasNeves: what do these square brackets do?

frontend/src/store/questions/index.js
```diff
@@ -51,15 +50,25 @@ const mutations = {
...
-    commit('REFRESH_FEED', await getQuestions())
-    commit('TOGGLE_LOADING', false)
+    try {
+      const data = await REQUEST[action](...args)
+      commit(COMMITS[action], data)
+    } catch (e) {
+      commit('API_ERROR', e)
+    } finally {
+      commit('TOGGLE_LOADING', false)
+    }
```
NunoDasNeves: okay I like this. It's what I was trying to do but better

frontend/src/utils/api/index.js
```diff
@@ -8,7 +8,7 @@ const API_URL = process && process.env ? 'http://localhost:3000/api' : 'data'
  * @returns {Promise}       The relevant request
  */
 export const get = (path, options = {}) =>
-  fetch(`${API_URL}${path}`, options)
+  fetch(`${API_URL}${path}`, {...options})
```
NunoDasNeves: Why?
options == {...options} right?

backend/src/models/db.js
DarkPurple141: Slightly confusing syntax I'll admit.

Leading semicolon so that the previous line doesn't execute the array. Reduce is passed two parameters, acc and curr. In this case every object in the array is assigned the user id.

frontend/src/store/questions/constants.js
DarkPurple141: yeap fair.

frontend/src/store/questions/index.js
DarkPurple141: yep

frontend/src/store/questions/constants.js
DarkPurple141: Allows the string to be dynamically set as the object key.

frontend/src/utils/api/index.js
DarkPurple141: yeah silly.

frontend/src/store/questions/index.js
DarkPurple141: Yeah it's nice ay.

backend/src/models/db.js
NunoDasNeves: If that's all it's doing, why not use map() or just for..of?

frontend/src/store/questions/constants.js
NunoDasNeves: neato

NunoDasNeves:
@DarkPurple141 fyi i jumped on here and removed everything in public folders and package-lock.json files, and added them to the .gitignore... it was bugging meeee

backend/src/models/db.js
DarkPurple141: Can't use .map() cos it returns a new array and has no side effects.

PR #120, opened 2018-09-02
S3 answer question backend

NunoDasNeves:


NunoDasNeves:
@DarkPurple141 senpai show me how to write a test for the controller/model or something

backend/src/models/answer.js
```diff
@@ -10,3 +10,16 @@ exports.getAnswers = function (questionID, pageNumber) {
...
+
+exports.postAnswer = function (questionID, {body}) {
+    return new Promise((resolve, reject) => {
+        const columns = ['userID', 'questionID', 'body']
+        const placeholders = columns.map(_ => '?').join()
+        const query = `INSERT INTO answer (${columns}) VALUES (${placeholders})`
+        db.run(
+            query,
+            [0, questionID, body], // TODO user id is a placeholder obviously, but it can't be null so...
+            function (err) { err ? reject(err) : resolve(err) }
```
PapaPwn: Resolving err?

NunoDasNeves: Actually, I want it to resolve getAnswers so it returns the updated list of answers...
This doesn't seem to work though `resolve(exports.getAnswers(questionID, 1))

NunoDasNeves: hmm yeah idk what's going on; i just get `TypeError: Cannot read property 'then' of undefined`

DarkPurple141:
I can’t seem to see these comments. Will have a good look later though.

backend/src/models/answer.js
```diff
@@ -10,3 +10,16 @@ exports.getAnswers = function (questionID, pageNumber) {
...
+
+exports.postAnswer = function (questionID, {body}) {
+    new Promise((resolve, reject) => {
+        const columns = ['userID', 'questionID', 'body']
+        const placeholders = columns.map(_ => '?').join()
+        const query = `INSERT INTO answer (${columns}) VALUES (${placeholders})`
+        db.run(
+            query,
+            [0, questionID, body], // TODO user id is a placeholder obviously, but it can't be null so...
+            function (err) { err ? reject(err) : resolve(exports.getAnswers(questionID, 1)) }
```
NunoDasNeves: @DarkPurple141 Basically I don't get what's going on here (this code doesn't work).
I wanna return getAnswers after adding the new answer to the db

DarkPurple141:
Found the comments. K having a look

backend/src/models/answer.js
DarkPurple141: Fixed with a missing return XD.

backend/src/controllers/question.js
```diff
@@ -13,3 +13,10 @@ exports.getQuestionAnswers = function ({ params, query }, res) {
     answerModel.getAnswers(params.id, query.p)
         .then(data => res.json(data))
 }
+
+/* POST new answer. */
+exports.postAnswer = function ({ params, body }, res) {
+    answerModel.postAnswer(params.id, body)
+        .then(data => res.json(data))
+        .catch(console.warn)
```
DarkPurple141: Not saying in this PR, but eventually just doing console.warns for real errors may not be the best idea.

frontend/src/store/questions/constants.js
```diff
@@ -1,23 +1,27 @@
...
+  [ACTIONS.GET_ANSWERS]: getAnswers,
+  [ACTIONS.POST_ANSWER]: postAnswer
 }
 
 export const COMMITS = {
   [ACTIONS.GET_QUESTIONS]: 'REFRESH_FEED',
   [ACTIONS.GET_QUESTION]: 'FOCUS_QUESTION',
-  [ACTIONS.GET_ANSWERS]: 'FOCUS_ANSWERS'
+  [ACTIONS.GET_ANSWERS]: 'FOCUS_ANSWERS',
+  [ACTIONS.POST_ANSWER]: 'FOCUS_ANSWERS'
```
DarkPurple141: Assuming this is intentional ie they both want to FOCUS_ANSWERS

frontend/src/utils/api/questions.js
```diff
@@ -49,10 +49,12 @@ export function getAnswers(id) {
 
 /**
  * @param {String} id the id of the question being answered
- * @param {object} body the data associated with the new answer {title, body}
+ * @param {object} body the data associated with the new answer
  */
-export function postAnswer(id, body) {
-  return post(`/question/${id}`, { body })
+export function postAnswer(id, form) {
+  const ctypeHeaders = new Headers();
```
DarkPurple141: if you call this variable headers, you can write the line 'headers' shortly after. ie

```js
return post(`/question/${id}/answers`, {headers, body:JSON.stringify(form)})
```

DarkPurple141: Although might need to change the way we do requests so we don't interfere with any existing headers. (Something for another day).

frontend/src/store/questions/index.js
```diff
@@ -43,9 +43,6 @@ const mutations = {
   API_ERROR (state, {code, message}) {
     state.error.code = code
     state.error.message = message
-  },
```
DarkPurple141: I suppose this makes sense to always rely on server as source of truth, seems kinda odd to be rerendering the whole list on a single question change tho.

frontend/src/store/questions/constants.js
```diff
@@ -1,23 +1,27 @@
 import {
   getQuestions,
   getQuestion,
-  getAnswers
+  getAnswers,
+  postAnswer
 } from '@/utils/api/questions'
 
 export const ACTIONS = {
```
DarkPurple141: Just a note on these symbols. They can (and probably should) be more descriptive as they're just unique key hashes but in debuggint they will give more relevant information. That was my bad for starting the convention.

frontend/public/index.html
```diff
@@ -0,0 +1,16 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
```
DarkPurple141: Eventually populate with more meta tags.

backend/src/models/answer.js
DarkPurple141: Sidenote I feel like calling things with exports.xxxx is sort of annoying.

See the way I made changes in db.js. Dunno!

frontend/src/store/questions/constants.js
NunoDasNeves: yep

PR #124, opened 2018-09-05
Playing with tests

Aruzal:
Played around with some testing.


Aruzal:
Not sure why test in tests/api.js for / is failing?

PapaPwn:
It's failing because /shizzlwazzle is returning status 200, not 404, which should supposedly be accounted for in the front end due to the comment:
```
/*
    anything that gets here and not handled
    by api error handler should get index.html.
    NB: frontend must handle actual 404s.
*/
app.use('*', function (_, res) {
    res.sendFile(path.join(__dirname, '../public', 'index.html'))
})
```
Not sure if the front end recognises that no data was returned from the request or if it works differently?

backend/tests/routes/api.js
```diff
@@ -15,6 +15,12 @@ describe('API route testing', () => {
             .expect(200)
     )
 
+    it('/api/bad/path', () =>
```
DarkPurple141: This one returns 404, but I think it was returning 200. I changed it.

backend/tests/routes/api.js
```diff
@@ -15,6 +15,12 @@ describe('API route testing', () => {
...
+        supertest
+            .get('/api/bad/path')
+            .expect(404)
+    )
+
     it('Test fallback for bad path', () =>
```
DarkPurple141: This should return 200.

DarkPurple141: (And it does)

backend/tests/routes/course.js
```diff
@@ -10,6 +11,48 @@ describe('Course route testing', () => {
...
+            .expect((req) => {
+                expect(req.body.courseID).to.be.a('number')
+                expect(req.body.courseCode).to.be.a('string')
+                expect(req.body.courseName).to.be.a('string')
+            })
+    )
+
+    describe('GET /api/course/1', () => {
+        let request
+
```
DarkPurple141: See the way I've done this section. Apply to the above section.

PR #133, opened 2018-09-07
S4 dependency injection models

DarkPurple141:
Changes that will make testing the models much more sensible by being able to stub the db.


backend/src/models/answer.js
```diff
@@ -8,20 +6,11 @@ class Answer {
...
-                [0, questionID, body], // TODO user id is a placeholder obviously, but it can't be null so...
-                // TODO meaningful error message/code based on sql error
-                // TODO proper error handling for all the endpoints :/
-                function (err) { err ? reject(Error('Error adding answer')) : resolve(self.getAnswers(questionID, 1)) }
-            )
-        })
+    postAnswer(questionID, { body, userID = 1 }) {
+        return this.db
+            .insert('answer', { questionID, body, userID })
+            /* Still not sure on this, seems wasteful to send all new data */
```
NunoDasNeves: I guess the alternative is selecting the answer we just inserted into the table and returning that. I think it's not worth the effort of adding right now though.

backend/src/models/answer.js
```diff
@@ -37,6 +26,16 @@ class Answer {
     }
 }
 
-module.exports = (function(db) {
+let Singleton = null
+
+/**
+ * @param {object} db defaults to the db instance
+ */
+module.exports = function(db) {
```
NunoDasNeves: this seems like a reasonable way to do it; defaulting to the actual database

backend/src/models/db/index.js
```diff
@@ -22,7 +22,7 @@ class DB {
     }
 
     deleteDB () {
-        return 0
+        return Promise.resolve(false)
```
NunoDasNeves: what's this doing?

backend/src/models/answer.js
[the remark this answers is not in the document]
DarkPurple141: is what you're saying:

```js
postAnswer(questionID, data)
```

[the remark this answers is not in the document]
DarkPurple141: or are you saying the following line.

backend/src/controllers/course.js
```diff
@@ -1,5 +1,5 @@
-const courseModel = require('../models/course')
-const questionModel = require('../models/question')
+const courseModel = require('../models/course')()
```
PapaPwn: So since we are not passing in a db, the test db is used, if I'm reading this correctly?

backend/src/models/question.js
```diff
@@ -35,10 +33,21 @@ class Question {
...
-module.exports = (function(db) {
+let Singleton = null
+
+/**
+ * @param {object} db defaults to the db instance
+ */
+module.exports = function(db) {
+    if (!db) {
+        /* app environment, dev or prod */
+        return (Singleton = Singleton ? Singleton : new Question(require('./db'))) // eslint-disable-line
```
PapaPwn: How is singleton set in prod?

backend/src/models/answer.js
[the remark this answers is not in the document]
PapaPwn: The following line, which the comment refers to:
```.then(() => this.getAnswers(questionID, 1))```

File: backend/tests/models/stub-db.js [hunk 6]
```diff
@@ -0,0 +1,19 @@
+class DB {
+    insert(table, data) {
```
NunoDasNeves: Isn't the idea of the dependency injection that we can restore the database to a known state before each test?
What's idea behind this stub?

backend/src/models/answer.js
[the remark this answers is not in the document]
DarkPurple141: You should click the line you're referring to then 👍

backend/src/controllers/course.js
DarkPurple141: when we are not passing in a db we use the singleton. Else we use the one we passed in.

backend/src/models/answer.js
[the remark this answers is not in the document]
PapaPwn: Wait dismiss this, you're referring to .insert to .then

DarkPurple141: Only thing I would say is that postQuestion and postAnswer currently behave differenlty.

And all the relevant data _except_ the id is already client side.

backend/src/models/answer.js
DarkPurple141: 👍

backend/src/models/db/index.js
DarkPurple141: Normalising the behaviour to reflect what the db does. Which is just return a promise. At the moment the data is nonsense, but might map the args back in future.

backend/src/models/question.js
DarkPurple141: Currently doesn't differentiate between prod/dev.

backend/src/models/answer.js
NunoDasNeves: yeah i spose

DarkPurple141: But, look I'll leave it for now.

PR #141, opened 2018-09-08
S4 ask questions

PapaPwn:
Just the basic ask question component. Tried to use https://github.com/froala/vue-froala-wysiwyg but couldn't figure out how 'webpack' works.


backend/src/models/question.js
```diff
@@ -35,6 +35,7 @@ class Question {
     postQuestion(courseID, { userID, title, body }) {
         return this.db
             .insert('question', { courseID, userID, title, body })
+            .then(() => this.getQuestions(courseID))
```
DarkPurple141: Still don't think we should do it like this. :(. But it's increasingly baked into the codebase.

frontend/src/components/questions-answers/QuestionForm.vue
```diff
@@ -0,0 +1,71 @@
...
+}
+</script>
+
+<style scoped lang='less'>
+
+.content {
+  padding: 0px 60px;
+}
+
+input[type=text] {
```
DarkPurple141: This should probably be a component.

frontend/src/views/Course.vue
```diff
@@ -4,12 +4,18 @@
       title="Latest Questions"
       :questions="feed"
     />
+    <QuestionForm @submitQuestionForm="submitQuestion" class="questionForm">
+      <span class="form-failure"
+        v-if="error.code">{{error.message}}
+      </span>
+    </QuestionForm>
   </div>
 </template>
```
DarkPurple141: Some of this may clash with Nuno's branch.

DarkPurple141:
Yes, render whenever no id prop is passed

frontend/src/router.js
```diff
@@ -30,6 +30,14 @@ export default new Router({
       // which is lazy-loaded when the route is visited.
       component: () => import('./views/Course')
     },
+    {
+      path: '/question',
+      name: 'newQuestion',
+      props: ({params: { id }}) => ({
```
DarkPurple141: This won't work. There are no params and yet you're passing one in?

DarkPurple141: Unless you mean query.

DarkPurple141:
@PapaPwn @NunoDasNeves There really needs to be a reset state function in the question store. IE you clear anything thats's in theory on some event.

NunoDasNeves:
The code looks okay... but try going directly to a question. e.g. http://localhost:3000/question?qid=1

PR #165, opened 2018-09-13
S5 backend pagination

Aruzal:
* Added pagination for questions and reviews (backend) using offtset and limit for sql queries.
* Added a simple check in the getQuestions and getReviews controller to make sure pageNumber is of type number (stops DB crashing for other weird stuff)


NunoDasNeves:
@Aruzal Good start - have you thought about the fact that multiple feeds get rendered on the Course page currently? I'm not sure how to deal with this... Maybe we need a more generic solution for paginating feeds? It might make sense to figure out the frontend for this first.
Please merge dev asap also - changes to api + routes will conflict with your changes somewhat

backend/src/controllers/course.js
```diff
@@ -18,13 +18,25 @@ exports.getCourse = function ({ params }, res) {
 
 /* Get all questions for a course */
 exports.getCourseQuestions = function ({ params, query }, res) {
-    responseHandler(questionModel.getQuestions(params.code, query.p), res)
+    // If query 'p' is not given just use p=1
+    let pageNumber = typeof query.p === 'undefined' ? 1 : parseInt(query.p)
+    // Check that pageNumber is an integer
```
DarkPurple141: assuming p = 0 is not a thing why not.

```js
// dunno
let p = parseInt(query.p)
const pageNumber = p ? p : 1
```

backend/src/controllers/course.js
```diff
@@ -18,13 +18,25 @@ exports.getCourse = function ({ params }, res) {
...
+    if (isNaN(pageNumber)) {
```
DarkPurple141: Means this is impossible.

backend/src/controllers/course.js
```diff
@@ -18,13 +18,25 @@ exports.getCourse = function ({ params }, res) {
...
         .catch(errorHandler(res))
 }
 
 /* Get all reviews for a course */
 exports.getCourseReviews = function ({ params, query }, res) {
-    responseHandler(reviewModel.getReviews(params.code, query.p), res)
+    // If query 'p' is not given just use p=1
+    let pageNumber = typeof query.p === 'undefined' ? 1 : parseInt(query.p)
+    // Check that pageNumber is an integer
+    if (isNaN(pageNumber)) {
```
DarkPurple141: Same deal.

backend/src/models/question.js
```diff
@@ -26,6 +15,19 @@ class Question {
...
+        let pageSize = 2
+        let offset = (pageSize * pageNumber) - pageSize
+        return this.db
+            .queryAll('SELECT * FROM question WHERE code=? ORDER BY timestamp DESC LIMIT ?, ?',
+                [code, offset, pageSize])
+    }
+
     /**
      * Post a question.
      * @param {string} code  The id from the route param
```
DarkPurple141: update this comment 👍

PR #166, opened 2018-09-16
S5 frontend pagination

Aruzal:
Not sure if this is the best way to do it but it works (suggested alternatives?)
Was not sure how to do pageNumber in the URL so I did it this way which works for now.

Known Issues
* Can increase pageNumber past max pages (since there is no max pages)

Still need to do pagination for Comments ect.


backend/src/controllers/course.js
```diff
@@ -18,13 +18,25 @@ exports.getCourse = function ({ params }, res) {
 
 /* Get all questions for a course */
 exports.getCourseQuestions = function ({ params, query }, res) {
-    responseHandler(questionModel.getQuestions(params.code, query.p), res)
+    // If query 'p' is not given just use p=1
+    let pageNumber = typeof query.p === 'undefined' ? 1 : parseInt(query.p)
+    // Check that pageNumber is an integer
+    if (isNaN(pageNumber)) {
+        throw Error('Page Number is not a number')
```
NunoDasNeves: What does this throw do currently? It should use the error handler to return some meaningful json to the frontend with error code 400.

backend/src/controllers/course.js
```diff
@@ -18,13 +18,25 @@ exports.getCourse = function ({ params }, res) {
...
 }
 
 /* Get all reviews for a course */
 exports.getCourseReviews = function ({ params, query }, res) {
-    responseHandler(reviewModel.getReviews(params.code, query.p), res)
+    // If query 'p' is not given just use p=1
+    let pageNumber = typeof query.p === 'undefined' ? 1 : parseInt(query.p)
+    // Check that pageNumber is an integer
+    if (isNaN(pageNumber)) {
+        throw Error('Page Number is not a number')
```
NunoDasNeves: ditto above

backend/src/models/question.js
```diff
@@ -26,6 +15,19 @@ class Question {
             .query('SELECT * FROM question WHERE id=?', [questionID])
     }
 
+    /**
+     * @param   {string} code        The code of the couse we're getting questions for.
+     * @param   {number} pageNumber  The pageNumber, defaults to 1?, if higher than max should just give max.
+     * @returns {object}
+     */
+    getQuestions(code, pageNumber) {
```
NunoDasNeves: Maybe a generic pagination interface could be used across models that query for multiple items from the database?

frontend/src/views/Course.vue
```diff
@@ -73,10 +85,51 @@ export default {
...
+        })
+    },
+    refreshReviews() {
+      this.$store.dispatch('course/getReviews',
+        {
+          id: this.code,
+          pageNumber: this.reviewPage
+        })
+    },
+    nextPage (feedType) {
```
NunoDasNeves: since you'll presumably extend Feed to work with answers, and update these functions according, you might want to make a function factory to generate nextPage/prevPage functions for answers, reviews and questions.

backend/src/models/question.js
Aruzal: I'm not really sure what you mean by this?

NunoDasNeves: I guess I meant having a function which structures paginated SQL queries for you... But now I realise it's probably out of scope of this issue; it should be done in a generic database layer if we end up adding that

backend/src/models/question.js
```diff
@@ -26,6 +15,19 @@ class Question {
...
+        let pageSize = 2
```
DarkPurple141: Almost all vars should be `const` unless otherwise necessary.

Aruzal: Noted and changed.

backend/src/controllers/course.js
```diff
@@ -20,15 +20,53 @@ exports.getCourse = function ({ params }, res) {
 exports.getCourseQuestions = function ({ params, query }, res) {
     let p = parseInt(query.p)
     const pageNumber = p || 1
-    responseHandler(questionModel.getQuestions(params.code, pageNumber), res)
+    const pageSize = 10
+
+    let getCourseQuestions = new Promise(function(resolve, reject) {
```
DarkPurple141: arrow func, and `const` over `let`

backend/src/controllers/course.js
```diff
@@ -20,15 +20,53 @@ exports.getCourse = function ({ params }, res) {
...
+        Promise.all([
+            questionModel.getQuestions(params.code, pageNumber, pageSize),
```
DarkPurple141: Oh. Shouldn't these be the same query with a nested sub query or does SQLite not allow that?

backend/src/controllers/course.js
```diff
@@ -20,15 +20,53 @@ exports.getCourse = function ({ params }, res) {
...
 /* Get all reviews for a course */
 exports.getCourseReviews = function ({ params, query }, res) {
     let p = parseInt(query.p)
     const pageNumber = p || 1
-    responseHandler(reviewModel.getReviews(params.code, pageNumber), res)
+    const pageSize = 10
+
+    let getCourseReviews = new Promise(function(resolve, reject) {
+        Promise.all([
+            reviewModel.getReviews(params.code, pageNumber, pageSize),
```
DarkPurple141: same as above.

backend/src/models/review.js
```diff
@@ -20,14 +20,24 @@ class Review {
...
             .queryAll('SELECT * FROM review WHERE code=? ORDER BY timestamp DESC LIMIT ?, ?',
                 [code, offset, pageSize])
     }
 
+    /**
+     * Gets the total number of reviews for a course
+     * @param   {string} code        The code of the course
+     * @returns {object}
+     */
+    getReviewCount(code) {
```
DarkPurple141: It would be maybe more understandable if this was separated if you only made the call on the initial getQuestions/getReviews call. But as we're returning with every request (which is also fine), this should be a subquery. Eg from databases.

```sql
select c.year, c.term, count(*), c.starting
from courses c
group by c.year, c.term, c.starting
order by c.starting
```

DarkPurple141: But maybe not possible in lite.

frontend/src/views/course/CourseQuestions.vue
```diff
@@ -15,34 +15,47 @@
         </ol>
       </section>
 
+      <PageSelector :currPage="this.meta.curr" :lastPage="this.meta.last" :update="refreshQuestions"></PageSelector>
```
DarkPurple141: `this` is redundant in the context of the template.

frontend/src/views/course/CourseQuestions.vue
```diff
@@ -15,34 +15,47 @@
...
   </div>
 </template>
 
 <script>
 // @ is an alias to /src
 import QuestionCard from '@/components/questions-answers/QuestionCard'
 import AppButton from '@/components/AppButton'
+import PageSelector from '@/components/PageSelector'
```
DarkPurple141: Generics should be prefaced with App.

DarkPurple141: `AppPageSelector`

backend/src/models/review.js
Aruzal: I tried doing something like this 
```
'SELECT q.title, q.body, COUNT(*) FROM question q WHERE code=? GROUP BY q.title, q.body ORDER BY timestamp DESC LIMIT ?, ?'
```
However COUNT is always 1, also It doesn't seem to work when using * to select all fields of the question. Might look at it later but the current solution works and is only slightly less efficient.

frontend/src/views/course/CourseReviews.vue
```diff
@@ -7,21 +7,21 @@
...
 
-      <section class="questions">
-        <ol>
-          <li :key="item.id" v-for="item in reviews">
-            <ReviewCard v-bind="item"/>
-          </li>
-        </ol>
-      </section>
+      <Feed
+        feedType="QuestionCard"
```
NunoDasNeves: ReviewCard

frontend/src/views/course/CourseReviews.vue
```diff
@@ -7,21 +7,21 @@
...
+        :items="questions"
```
NunoDasNeves: reviews

frontend/src/views/course/CourseReviews.vue
```diff
@@ -7,21 +7,21 @@
...
+      />
+
+    <AppPageSelector :currPage="meta.curr" :lastPage="meta.last" :update="refreshReviews"></AppPageSelector>
```
NunoDasNeves: v-if="meta.last != 1"

frontend/src/views/course/CourseQuestions.vue
```diff
@@ -7,42 +7,52 @@
...
-            <QuestionCard v-bind="item"/>
-          </li>
-        </ol>
-      </section>
+      <Feed
+        feedType="QuestionCard"
+        :items="questions"
+      />
 
+      <AppPageSelector :currPage="meta.curr" :lastPage="meta.last" :update="refreshQuestions"></AppPageSelector>
```
NunoDasNeves: v-if="meta.last != 1"

frontend/src/views/course/CourseReviews.vue
Aruzal: Good catch

PR #173, opened 2018-09-21
S5 breadcrumb

NunoDasNeves:
I think this is good enough for now.
Worth adding to once we have backend and frontend for pages listing all Subjects (COMP, ACCT etc) and listing Courses by Subject.

### DIFF
--- a/frontend/src/components/Card.vue
+++ b/frontend/src/components/Card.vue
@@ -7,7 +7,7 @@
 <style scoped>
 .card {
   position: relative;
-  margin: 10px 0px;
+  margin-bottom: 10px;
   border: var(--border);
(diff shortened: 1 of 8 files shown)


frontend/src/components/Card.vue
```diff
@@ -7,7 +7,7 @@
 <style scoped>
 .card {
   position: relative;
-  margin: 10px 0px;
```
DarkPurple141: This affects multiple components are you sure this is non breaking?

NunoDasNeves: Yeah I mean I've tested it everywhere card is used.

DarkPurple141:
✅

PR #194, opened 2018-10-03
S6 better reviews

NunoDasNeves:
Some things need fixing/improving like the text on the AppRatingCircles.

Feedback  on review style would be good. I think reviews should be kinda like this, with some ratings optional and others we force people to give.

Aggregation isn't quiite working properly for small numbers of reviews due to integer division I think, but works fine in general I think.

I might have missed some stuff...

TODO in a different issue/branch is displaying the ratings of a given review in the review feed/single review page.


backend/src/models/review.js
```diff
@@ -43,10 +45,21 @@ class Review {
      * @param {object} data  controller passed in object which should
      *                       contain the user data (probs eventually from an auth token)
      */
-    postReview(code, { title, body, userID = 1 }) {
+    postReview(code, { title, body, recommend, enjoy, difficulty, teaching, workload, userID = 1 }) {
+
+        if (recommend != 0 && recommend != 1) throw Error("Invalid recommend value")
+        if (enjoy < 1 || enjoy > 5) throw Error("Invalid enjoy value")
+
+        ;[difficulty, teaching, workload].forEach(item => {
```
Aruzal: What does the ; do?

frontend/src/utils/api/reviews.js
```diff
@@ -28,6 +28,18 @@ export function reviewMapper({ id, code, likes, userID, title, body, timestamp }
   }
 }
 
+export function newReviewMapper({ title, body, recommend, enjoy, difficulty, teaching, workload }) {
+  return {
+    title,
+    body,
+    recommend: {'Yes': 1, 'No': 0}[recommend],
```
Aruzal: Confused about the {}[] syntax, is this matching the variable in [] to a key in the {} to set the value?

frontend/src/components/AppRatingCircle/index.vue
```diff
@@ -23,25 +23,16 @@ export default {
...
   },
   computed: {
     circumference() {
       return 2 * Math.PI * this.r
-    }
-  },
-  methods: {
-    calcRating() {
+    },
+    rating() {
```
DarkPurple141: transition will only work with a starting value. I think the starting value needs to be this.circumfrence.

So the place it looks to rating it should instead default to rating || circumfrence.

Speaking of, given that value is undefined this will also run into errors without a check.

frontend/src/components/reviews-replies/ReviewForm.vue
```diff
@@ -8,22 +8,32 @@
...
 
-            <InputOptions v-model="recommend" :options="['Yes', 'No', 'Unsure']">
-              Would you recommend this course?
+            <InputOptions v-model="recommend" :options="['Yes', 'No']">
+              Would you recommend this course to a friend? <span class="required">*</span>
             </InputOptions>
 
-            <InputOptions v-model="enjoy" :options="['Yes', 'No', 'Unsure']">
-              Did you enjoy this course?
+            <InputOptions v-model="enjoy" :options="['1', '2', '3', '4', '5']">
```
DarkPurple141: Still unsure on numbers. I think simplified choices are generally best.

frontend/src/components/reviews-replies/ReviewOptions.vue
```diff
@@ -39,7 +39,7 @@ export default {
 
 .options {
     display: grid;
-    grid-template-columns: 1fr 1fr 1fr;
+    grid-template-columns: 1fr 1fr 1fr 1fr 1fr;
```
DarkPurple141: figured out the way I meant was:

```css
.options {
  grid-auto-columns: 1fr;
}

.option {
  grid-row: 1;
}
```

frontend/src/store/course/index.js
```diff
@@ -26,6 +25,13 @@ const state = {
 
 const getters = {
   course: ({course}) => course,
+  ratings: ({course}) => [
+    {text: 'Recommended', value: course.recommend || 0},
```
DarkPurple141: Should null values resolve. I guess this safe, but it also creates a problem where courses get 0 ratings if unrated, which I'm not sure on.

DarkPurple141:
Looks good. Just the grid change.

DarkPurple141:
Bug?

screen shot 2018-10-03 at 3 35 21 pm

DarkPurple141:
I think keep all compulsory, and make enjoy Yes, No, Meh.

That's my two cents.

frontend/src/components/reviews-replies/ReviewForm.vue
NunoDasNeves: I'm not 100% sure either...did this on a whim.
But I actually think mixing up the method of input is worthwhile.
I think we should discuss tomorrow and figure out exactly what to put on reviews for now.

frontend/src/components/reviews-replies/ReviewOptions.vue
NunoDasNeves: Yes this works! Will amend

frontend/src/store/course/index.js
NunoDasNeves: Yeah ... maybe a greyed out circle with no number or 'unrated' text would be better. Feel free to add if you see a good way to do it... i'm outta time today

NunoDasNeves:
@DarkPurple141 I'm not sure about that bug... it looks like tables haven't been initialized properly, doesn't seem to happen for me. Is it always happening?

DarkPurple141:
@NunoDasNeves Happening on Staging Consistently when creating a new review.

PR #209, opened 2018-10-04
S7 upvotes

PapaPwn:
Upvotes and downvotes are implemented on the backend and used minimalistically on the frontend.


backend/src/controllers/question.js
```diff
@@ -1,22 +1,71 @@
...
 const { responseHandler } = require('../utils/helpers')
 
 /* GET question data. */
 exports.getQuestion = function ({ params }, res) {
-    responseHandler(questionModel.getQuestion(params.id), res)
+    const getQuestion = Promise.all([
+        questionModel.getQuestion(params.id),
+        likesModel.getLikes({ type: 'question', id: params.id })
+    ])
+        .then(([question, likes]) => { return { ...question, ...likes } })
```
DarkPurple141: This will spread the object into a flat structure. What does the likes object look like?

backend/src/models/db/tables.js
```diff
@@ -211,8 +229,55 @@ function insertDB (db, table, data, prep) {
...
+}
+
+// Updates given JSON object into database table.
+// data = { column : value }
+// For security reasons, column inputs can NEVER be user defined.
+// TODO - Maybe need a better way todo more complex conditions
+function updateDB (db, table, data, conditions) {
+    return new Promise((resolve, reject) => {
+        const updateValues = Object.values(data)
+        const updateColumns = Object.keys(data)
```
DarkPurple141:
```js
const [ columns, values ] = Object.entries(data)
```

backend/src/routes/question.js
```diff
@@ -11,4 +11,16 @@ question.get('/:id/answers', questionController.getQuestionAnswers)
 /* Post an answer for a given question */
 question.post('/:id/answers', questionController.postAnswer)
 
+/* Get the question like value */
```
DarkPurple141: Not using auth yet?

frontend/src/components/comments/CommentCard.vue
```diff
@@ -20,11 +16,37 @@
...
+  components: { Card, User, Vote },
   props: {
+    type: String,
+    id: String,
+    code: String,
     comment: Object
+  },
+  methods: {
+    upvote() {
+      // TODO - REMOVE USER ID FROM HERE
```
DarkPurple141:
```
const { type, code, id, comment } = this
```

frontend/src/store/questions/index.js
```diff
@@ -37,7 +37,11 @@ const mutations = {
   FOCUS_ANSWERS (state, answers) {
     state.questionObj.answers = answers.map(answerMapper)
   },
-  API_ERROR (state, {code, message}) {
+  FOCUS_LIKES (state, { likes }) {
+    state.questionObj.question.likes = likes
```
DarkPurple141: Can be handled in frontend logic I think.

backend/src/controllers/question.js
PapaPwn: Just a normal question object with the addition of likes

backend/src/models/db/tables.js
PapaPwn: This didn't work for some reason, didn't waste time looking into it too much (the columns and values are not the same as what the above code produces)

frontend/src/store/questions/index.js
PapaPwn: When you play around with this particular component to make it pretty feel free to transfer the logic, for now this is just a simple way to ensure you aren't doubling up on votes.

PR #212, opened 2018-10-09
S8 db dev data

Aruzal:
Not sure if it should be called like this as we may not want to populate with test data?


backend/src/models/db/tables.js
```diff
@@ -179,6 +179,195 @@ function initCourseTable(db) {
...
+    ]
+
+    const minRange = 10 // Between [minRange, minRange+maxRange]
+    const maxRange = 5
+    const numQuestionsTypes = questionTypes.length
+
+    let questions = []
+
+    // For each of the courses
+    for (const i in courseData) {
```
DarkPurple141: if you find yourself getting an index which you're then just using to access the array, you should always opt for:

```js
for (const item of courseData) {

}
```

Or use map.

PR #219, opened 2018-10-11
S8 Card Feeds

DarkPurple141:
Card Feed updates.


frontend/src/components/AppBreadCrumb.vue
```diff
@@ -54,7 +54,8 @@ export default {
         }
       }
 
-      return list
+      // only last three items
```
PapaPwn: I think should be the last 4 items, but not including the most recent if you only want 3 on the screen. Reasoning being:
- like you mentioned the last being redundant
- it is still possible that from a question/review that I may want to quickly browse back to the subjects page.

DarkPurple141: Yep fair enuf.

PR #229, opened 2018-10-14
Fixed feed likes

PapaPwn:
Not the cleanest but works.

### DIFF
--- a/backend/web.config
+++ b/backend/web.config
@@ -35,7 +35,11 @@
         </rule>
       </rules>
     </rewrite>
-    
+
+    <staticContent>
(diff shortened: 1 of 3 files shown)


backend/web.config
```diff
@@ -35,7 +35,11 @@
         </rule>
       </rules>
     </rewrite>
-    
+
+    <staticContent>
+        <mimeMap fileExtension="." mimeType="text/plain" />
```
PapaPwn: This was just a thing I needed to host files without an extension so that we could prove identity to get an SSL certificate.

PapaPwn: No longer needed, but good for renewals.

PR #232, opened 2018-10-15
S8 restrict api

PapaPwn:
Restrict the backend too, plus some more login redirection, plus a couple of bug fixes.
AND
Make the frontend respond with prompts to login.


backend/package.json
```diff
@@ -16,6 +16,7 @@
     "express": "~4.16.0",
     "firebase-admin": "^6.0.0",
     "morgan": "~1.9.0",
+    "node-fetch": "^2.2.0",
```
DarkPurple141: This is a dev dependency, not a application dependency. We only use it for testing.

backend/src/utils/cors.js
```diff
@@ -18,7 +18,7 @@ exports.cors = function(req, res, next) {
 }
 
 exports.corsProd = function(req, res, next) {
-    res.header('Access-Control-Allow-Origin', 'https://smartcourse.me')
+    res.header('Access-Control-Allow-Origin', 'https://smartcourse.me, https://wwww.smartcourse.me')
```
DarkPurple141: I'm fairly sure you can't specify multiple origin in this header. The only way to deal with this is to dynamically update the header at runtime depending on the request.

frontend/src/views/Review.vue
```diff
@@ -53,6 +53,12 @@ export default {
   },
   methods: {
     submitReply (replyForm) {
+      const authState = this.$store.getters['auth/isLoggedIn']
```
DarkPurple141: Don't do this. We're going to remove the functionality. And prompt, not force them.

DarkPurple141: Same with above.

frontend/src/components/reviews-replies/ReviewCard.vue
```diff
@@ -37,11 +37,21 @@ export default {
   },
   methods: {
     upvote() {
+      const authState = this.$store.getters['auth/isLoggedIn']
```
DarkPurple141: Same with this. I think better to prompt user then force redirect.

backend/src/utils/cors.js
PapaPwn: Yeah you are right, will fix this.

frontend/src/views/Review.vue
PapaPwn: Do we have hidden prompt functionality or should I implement this now?

DarkPurple141: There's not presently. I was thinking button would be replaced with text if user is not authenticated. "You need to be logged in to write a review" etc OR button goes to disabled state and has a hover text.

PapaPwn: That's a lot of text for a button. The second option might be viable. I'll have a play and see what feels natural :)

DarkPurple141: It wouldn't be a button.

NunoDasNeves: ^

frontend/src/views/Question.vue
```diff
@@ -1,16 +1,16 @@
 <template>
     <section class="main-content">
       <AppBreadCrumb/>
-      <QuestionCard v-bind="question"/>
+      <QuestionCard v-bind="question" :authenticated="authenticated"/>
 
-      <AnswerForm @submitCommentForm="submitAnswer" :type="commentType" :callback="submitAnswer">
+      <AnswerForm @submitCommentForm="submitAnswer" :type="commentType" :callback="submitAnswer" :authenticated="authenticated">
```
NunoDasNeves: For the forms I think it'd be better if we replaced the whole form with text (e.g. "Login to answer this question") or ifwe greyed out/disabled the whole form

DarkPurple141: Agree with the first part. I think the former just disappears and becomes a sentence of text.

PapaPwn: No problemo, but making things pretty isn't a strength of mine, so I might leave the CSS to @DarkPurple141 , but the functionality will be there.

PapaPwn: Actually the functionality is already there you just need to add `v-if="authenticated"` and a `v-else` :)

DarkPurple141:
I'll be interested to know how much that v-tooltip adds to the bundle size.

frontend/src/components/AppButton.vue
```diff
@@ -1,7 +1,16 @@
 <template>
-    <button type='button' :class="alt ? 'standard-button' : 'alt-button'">
-        <slot></slot>
-    </button>
+    <div>
```
DarkPurple141: This seems like two components.  Else, one component but oddly shaped.

frontend/src/components/Comments/CommentForm.vue
```diff
@@ -2,7 +2,9 @@
     <CardForm :title="`Submit ${type}`">
       <textarea placeholder="Your input here.." v-model="body"></textarea><br>
       <!-- should probs be a separate component -->
-      <AppButton @click.native="callback({body}); body=''">{{ type }}</AppButton>
+      <AppButton :disabled="!(authenticated)" :disabledMsg="disabledMsg" @click.native="callback({body}); body=''">
```
DarkPurple141: brackets?

PapaPwn: No reason Mr Linter.

frontend/src/components/AppButton.vue
PapaPwn: It's one and oddly shaped, just wanted something that worked. Not sure about the proper pattern here.

PapaPwn:
Going to merge this in and the style changes can be made as a separate PR.

PR #236, opened 2018-10-16
S9 loading state

NunoDasNeves:
Loading state pretty much everywhere its needed now.
Some other optimizations to reduce the number of times courses get loaded from api.
Might have gone overboard on that buuut I think it's ok. Thoughts?

### DIFF
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -150,4 +150,12 @@ textarea:focus, textarea:active {
   border: 1px solid #acc;
 }
 
+.fade-slide-enter-active {
+  transition: all 0.5s;
+}
+.fade-slide-enter {
+  opacity: 0;
+  transform:translateY(10px);
+}
+
 </style>
--- a/frontend/src/App.vue
+++ b/frontend/src/App.vue
@@ -151,7 +151,7 @@ textarea:focus, textarea:active {
 }
 
 .fade-slide-enter-active {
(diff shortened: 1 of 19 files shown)


frontend/src/store/root.js
```diff
@@ -1,37 +1,54 @@
...
 }
 
 const mutations = {
-  SET_LOADING(state, bool) {
+  TOGGLE_LOADING(state, bool) {
     state.loading = bool
   },
-  POPULATE_SEARCH(state, data) {
-    state.courses = data
+  REFRESH_COURSES(state, courses) {
```
DarkPurple141: This is for what, a review update?

NunoDasNeves: Yeah, atm the idea is to load courses once only then update whenever an individual course is loaded...

NunoDasNeves:
@darkpurple141 added transitions. I think it's marginally better ... Not sure what else to do

NunoDasNeves:
Oh there's conflicts...

frontend/src/App.vue
```diff
@@ -149,4 +150,12 @@ textarea:focus, textarea:active {
   border: 1px solid #acc;
 }
 
+.fade-slide-enter-active {
+  transition: all 0.5s;
```
DarkPurple141: transition: prop time ease-in-out, prop time ease-in-out;

PR #241, opened 2018-10-17
S9 question card num answers

NunoDasNeves:
Question cards in Feed now show the number of answers
And I added likes to the dummy data
And enriched the dummy data a bit
And disabled like buttons when not logged in. As in actually disabled them so clicking does nothing
Annnd it won't say -1 users found this helpful on the feed anymore
yaayyy


backend/src/models/db/tables.js
```diff
@@ -211,18 +211,19 @@ function initQuestionsTable(db) {
...
-    const minRange = 10 // Between [minRange, minRange+maxRange]
-    const maxRange = 5
+    const minRange = -2 // Between [minRange, minRange+maxRange]
+    const maxRange = 15
     const numQuestionsTypes = questionTypes.length
 
     let questions = []
 
     // For each of the courses
     for (const i in courseData) {
```
DarkPurple141: persible bug.

frontend/src/components/Course/FeedCard.vue
```diff
@@ -29,6 +30,7 @@ export default {
     likes: Number,
     published: String,
     user: Object,
+    numAnswers: {Type: Number, Default: 0},
```
DarkPurple141: loweercase

PR #242, opened 2018-10-17
very basic profiles without attempting to make anything pretty

NunoDasNeves:
title


frontend/src/components/User/Mini.vue
```diff
@@ -1,11 +1,11 @@
 <template>
-    <div class="avatar">
+    <router-link :to="{ name: 'Profile', params: { id } }" tag='div' class="avatar">
```
DarkPurple141: Yeah I meant to do this previously.

PR #246, opened 2018-10-18
Faculty degree data

NunoDasNeves:
degree + faculty shiz


backend/data/degrees.js
```diff
@@ -0,0 +1,2671 @@
+module.exports = [
```
DarkPurple141: Should we just include the prefiltered data?

NunoDasNeves: myeh...

PR #247, opened 2018-10-18
S9 Comment Feed Rework

NunoDasNeves:
Just answers (no review replies) so far, will add to this PR tomorrow I guess...

Idk how to do a good animation for hiding/showing the answer thing. `flex-direction` isn't transition-able.

I fixed some backend jank too; postAnswer returns just the answer you posted now, which gets unshifted onto the answer array so it appears at the top.

Sorting of answers by 'hotness' should be applicable to anything with likes and timestamp. Should probably factor user reputation into that too...

UPDATE:
bugfixes, code cleanup, reply section rework
animation didn't look good, even with `height`

### DIFF
--- a/frontend/src/components/Course/Controls.vue
+++ b/frontend/src/components/Course/Controls.vue
@@ -28,13 +28,11 @@ export default {
   computed: {
     authenticated() {
       return this.$store.getters['auth/isLoggedIn']
-    }
-  },
-  data() {
-    return {
-      disabledMessage: {
-        content: 'You must be logged in to post.',
-        placement: 'left'
+    },
+    disabledMessage() {
+      return {
+          content: this.authenticated ? '' : 'You must be logged in to post.',
+          placement: 'left'
       }
     }
   }


DarkPurple141:
Visually looks good man nice job.

backend/src/controllers/course.js
```diff
@@ -90,14 +90,14 @@ exports.getCourseReviews = function ({ params, query }, res) {
...
-    body.userID = user && user.id || ANONYMOUS
+    body.userID = user.id
     responseHandler(questionModel.postQuestion(params.code, body), res)
         .catch(errorHandler(res))
 }
 
 /* POST new review */
 exports.postReview = function ({ user, params, body }, res) {
-    body.userID = user && user.id || ANONYMOUS
+    body.userID = user.id
```
DarkPurple141: Good. Because you can no longer get here without a userID

backend/src/models/comment.js
```diff
@@ -15,8 +15,7 @@ class Comment {
         const [key, value] = Object.entries(queryObject)[0]
         return this.db
             .insert('comment', { [key]: value, body, userID })
-            /* Still not sure on this, seems wasteful to send all new data */
-            .then((id) => ({ body, userID, id, timestamp: Date.now() }))
+            .then((id) => this.getComment(id))
```
DarkPurple141: Ahh because otherwise we didn't get all fields, we just got the id.

frontend/src/components/Comments/CommentForm.vue
```diff
@@ -1,41 +1,68 @@
 <template>
-    <CardForm :title="`Submit ${type}`">
```
DarkPurple141: Does CardForm get used anywhere else? Maybe that component should be reworked to the below.

frontend/src/components/Comments/CommentForm.vue
```diff
@@ -1,41 +1,68 @@
...
-    authenticated: Boolean
+    authenticated: Boolean,
+    closeCallback: Function
   },
   data () {
     return {
       body: '',
       disabledMessage: {
-        content: 'You must be to logged in for this functionality.',
+        content: 'You must be logged in for this functionality.',
```
DarkPurple141: note, if you are logged in this message still displays on hover so may need to be dynamic.

frontend/src/components/Comments/CommentForm.vue
```diff
@@ -1,41 +1,68 @@
...
         placement: 'right'
       }
     }
   }
 }
 </script>
+
+<style scoped>
+h4 {
+  font-weight: normal;
```
DarkPurple141: we should amend the font-matrix if there's nothing that matches desired size.

frontend/src/components/Comments/CommentSpacer.vue
```diff
@@ -0,0 +1,11 @@
+<template>
+  <div class="bar"><slot/></div>
```
DarkPurple141: new line for the slot would be good just so it's easy to see what's going on.

frontend/src/store/utils.js
```diff
@@ -2,14 +2,35 @@
 // factories
 export function doRequestFactory(REQUEST, COMMITS) {
   return async function ({ commit }, { action, args, load = 'TOGGLE_LOADING' }) {
```
DarkPurple141: load has a default thought so won't this always resolve?

backend/src/models/comment.js
NunoDasNeves: Yes this was actually broken. Only reason it didn't break frontend was because of another bug I think

frontend/src/components/Comments/CommentForm.vue
NunoDasNeves: Workin' on it

frontend/src/store/utils.js
NunoDasNeves: no cos we can pass `''` or `null` which are falsey and won't be overridden by the default

NunoDasNeves: actually i'm not 100% sure about `null` but `''` definitely works

frontend/src/components/Comments/CommentForm.vue
NunoDasNeves: So I've ended up just using Card here... I think CardForm looks good for question & review still, so I don't want to change it.
I could have made CardFormSmall or something... but don't think it's worth it rn

frontend/src/components/Comments/CommentForm.vue
NunoDasNeves: dynamised

frontend/src/components/Comments/CommentForm.vue
NunoDasNeves: body-copy-1 works

frontend/src/components/Comments/CommentSpacer.vue
NunoDasNeves: donezo

NunoDasNeves:
@DarkPurple141 reviews done, i think is good now

frontend/src/components/Comments/CommentForm.vue
DarkPurple141: Oh yeah I forgot for NewView.

PR #252, opened 2018-10-19
S9 unlikes

PapaPwn:
Still has the bug described in slack, but to be fixed l8rs.


frontend/src/store/questions/index.js
```diff
@@ -60,8 +60,9 @@ const mutations = {
     state.questionObj.answers = [ans, ...ordered]
   },
   FOCUS_LIKES (state, { likes }) {
+    const oldLikes = state.questionObj.question.likes
     state.questionObj.question.likes = likes
-    // TODO - ALREADY UPVOTED
+    state.questionObj.question.userLiked += likes - oldLikes
```
NunoDasNeves: I thought we got userLikes in the object anyway?
This method is prone to races with other users isn't it?

frontend/src/store/reviews/index.js
```diff
@@ -43,8 +43,9 @@ const mutations = {
     state.reviewObj.replies = sortByHotness(replies.map(replyMapper))
   },
   FOCUS_LIKES (state, { likes }) {
+    const oldLikes = state.reviewObj.review.likes
     state.reviewObj.review.likes = likes
-    // TODO - ALREADY UPVOTED
+    state.reviewObj.review.userLiked += likes - oldLikes
```
NunoDasNeves: ^ see above

frontend/src/store/questions/index.js
PapaPwn: Doing a PUT request for likes, just returns the new "likes" value for the object. userLiked is not userLikes, it is what the current user has voted (-1, 0, +1). However it is prone to a race. Perhaps I'll just return the correct "userLiked" value with the "likes" value?

backend/src/controllers/question.js
```diff
@@ -6,16 +7,18 @@ const errorHandler = require('./error')
...
 /* GET question data. */
-exports.getQuestion = function ({ params }, res) {
+exports.getQuestion = function ({ user, params }, res) {
+    const userID = user && user.id || ANONYMOUS
     const getQuestion = Promise.all([
         questionModel.getQuestion(params.id),
-        likesModel.getLikes({ type: 'question', id: params.id })
-    ]).then(([question, likes]) => {
+        likesModel.getLikes({ type: 'question', id: params.id }),
+        likesModel.getUserLiked({ type: 'question', id: params.id, userID })
```
DarkPurple141: I feel like this is an unneeded additional call if we just got return list of likes in the `.getLikes` call. Then you can deduce the second part trivially.

backend/src/controllers/question.js
```diff
@@ -24,20 +27,26 @@ exports.getQuestion = function ({ params }, res) {
...
         // Get the answers
         commentModel.getComments({ questionID: params.id }, query.p)
             .then(answers => Promise.all([
                 answers,
                 Promise.all(
-                    answers.map(answer => likesModel.getLikes({ type: 'answer', id: answer.id }))
+                    answers.map(answer => likesModel.getLikes(
+                        { type: 'answer', id: answer.id }))
+                ),
+                Promise.all(
```
DarkPurple141: Can get rid of this if you do that.

backend/src/controllers/review.js
```diff
@@ -6,16 +7,18 @@ const errorHandler = require('./error')
...
+exports.getReview = function ({ user, params }, res) {
+    const userID = user && user.id || ANONYMOUS
     const getReview = Promise.all([
         reviewModel.getReview(params.id),
-        likesModel.getLikes({ type: 'review', id: params.id })
-    ]).then(([review, likes]) => {
+        likesModel.getLikes({ type: 'review', id: params.id }),
+        likesModel.getUserLiked({ type: 'review', id: params.id, userID })
+    ]).then(([review, likes, userLiked]) => {
         return Promise.all([
```
DarkPurple141: unnecssary Promise.all()

backend/src/controllers/review.js
```diff
@@ -24,19 +27,26 @@ exports.getReview = function ({ params }, res) {
...
+        // Get the replies
         commentModel.getComments({ reviewID: params.id }, query.p)
             .then(replies => Promise.all([
                 replies,
                 Promise.all(
-                    replies.map(reply => likesModel.getLikes({ type: 'reply', id: reply.id }))
+                    replies.map(reply => likesModel.getLikes(
+                        { type: 'reply', id: reply.id }))
+                ),
+                Promise.all(
```
DarkPurple141: Still feel like this is redundant.

frontend/src/utils/api/index.js
```diff
@@ -21,11 +21,6 @@ async function responseCheck(res) {
 function request (path, { headers, method, data }) {
   // eventually add cors and auth headers
   const url = `${API_URL}${path}`
-
-  if (method === 'GET') {
```
DarkPurple141: I still don't think sending the auth token with every req is good.

PapaPwn: I agree, but for now I think its okay

backend/src/controllers/review.js
PapaPwn: Resolved in reputation PR

frontend/src/store/questions/index.js
PapaPwn: Will address if I have time though

PR #253, opened 2018-10-19
S9 faculty degree

Aruzal:


backend/src/controllers/uni.js
```diff
@@ -5,24 +5,28 @@ const { responseHandler } = require('../utils/helpers')
...
 /* Get all faculties */
 exports.getFaculties = function (_, res) {
+    res.header('maxAge', '365d')
     responseHandler(uniModel.getFaculties(), res)
         .catch(errorHandler(res))
 }
 
 /* Get all courses of a given subject */
 exports.getCourses = function ({ params }, res) {
+    res.header('maxAge', '365d')
```
DarkPurple141: Might want to check this works Trav? Can you make sure the response from the API has the correct headers? Should be Cache-Control: .... in the response.

frontend/src/store/root.js
```diff
@@ -27,6 +31,28 @@ const actions = {
       .then(data => commit('REFRESH_COURSES', data))
       .catch(err => console.warn(err))
       .finally(() => commit('TOGGLE_LOADING', false))
+  },
+  getFaculties({ commit, getters }) {
+    // avoid repeats of this
+    if (getters.faculties.length) {
+      return
+    }
+    commit('TOGGLE_LOADING', true)
```
DarkPurple141: Wouldn't even bother with toggling loading. As this has to be loaded at App init. Keep it simple basically.

DarkPurple141:
@Aruzal don't worry about simplifying the filters, but just make sure the headers are definitely working.

PR #255, opened 2018-10-19
S9 reputation

PapaPwn:
What even is a database.


backend/src/models/user.js
```diff
@@ -21,8 +25,11 @@ class User {
      * @returns {object}
      */
     getPublicProfile(id) {
-        return this.db
-            .query('SELECT id, displayName, degree, gradYear, description, picture, reputation, joined FROM user WHERE id=?', [id])
+        return this.db.query('SELECT id, displayName, degree, gradYear, description, picture, reputation, joined FROM user WHERE id=?', [id])
+            .then((profile) => {
+                if (profile.reputation < 0) profile.reputation = 0
+                return { ...profile }
```
DarkPurple141: this should just be return profile. Spread is redundant.

backend/src/models/user.js
```diff
@@ -40,6 +47,9 @@ class User {
      *                       contain the user data (probs eventually from an auth token)
      */
     createUser({ displayName, email, uid }) {
```
DarkPurple141: Isn't this called after we hit the firebase api? This isn't as trivial as changing this field, because the email, uid will be potentially created by firebase even if our backend rejects. I'm going to look at this today and tomorrow in any case.

backend/src/utils/helpers.js
```diff
@@ -42,7 +42,7 @@ exports.userLikesMapper = (likes, userLikes) => (
     user: {
         id: userID,
         displayName,
-        reputation,
+        ...(reputation >= 0 ? { reputation } : { 'reputation': 0 }),
```
DarkPurple141: or just:

```
reputation: (reputation >= 0 ? reputation: 0), 
```

backend/src/models/user.js
```diff
@@ -9,9 +9,13 @@ class User {
...
      * @param {string} id The id of the auth'd user
      */
+
     getProfile(id) {
-        return this.db
-            .query('SELECT id, email, displayName, degree, gradYear, description, picture, reputation, joined FROM user WHERE id=?', [id])
+        return this.db.query('SELECT id, email, displayName, degree, gradYear, description, picture, reputation, joined FROM user WHERE id=?', [id])
+            .then((profile) => {
+                if (profile.reputation < 0) profile.reputation = 0
+                return { ...profile }
```
DarkPurple141: should just return profile.

backend/src/models/likes.js
```diff
@@ -34,13 +34,46 @@ class Likes {
...
+
+        // Stuff to update likes
+        const data = {
+            objectType: type,
+            objectID: id,
+            userID,
+            value
+        }
+        const insertValues = Object.values(data)
+        const insertColumns = Object.keys(data)
```
DarkPurple141: const [columns, values] = Object.entries(data)... instead of Keys and values.

backend/src/controllers/question.js
```diff
@@ -14,12 +14,11 @@ exports.getQuestion = function ({ user, params }, res) {
         likesModel.getLikes({ type: 'question', id: params.id }),
         likesModel.getUserLiked({ type: 'question', id: params.id, userID })
     ]).then(([question, likes, userLiked]) => {
-        return Promise.all([
-            userModel.getPublicProfile(question.userID)
-        ]).then(([userInfo]) => {
-            delete question.userID
-            return { ...question, ...likes, ...userLiked, user: userInfo }
-        })
+        return userModel.getPublicProfile(question.userID)
```
DarkPurple141: This doesn't need to be nesting deeper and the return is redundant if u take out the curly bracket.

backend/src/models/likes.js
```diff
@@ -34,13 +34,46 @@ class Likes {
...
+        const insertPlaceholders = insertColumns.map(_ => '?').join()
+        const updateLikes = `REPLACE INTO likes (${insertColumns}) VALUES (${insertPlaceholders})`
+
+        // Stuff to update reputation
+        const creatorTable = (type === 'question' || type === 'review') ? type : 'comment'
+        const updateReputation = `UPDATE user
+        SET reputation = (SELECT reputation FROM user WHERE id = ?) + ?
```
NunoDasNeves: pretty sure this can just be `SET reputation = reputation + ?`

backend/src/models/likes.js
```diff
@@ -34,13 +34,46 @@ class Likes {
...
+                this.db.query('SELECT * FROM likes WHERE objectType=? AND objectID=? AND userID=?',
+                    [type, id, userID]),
+                this.db.query(`SELECT userID FROM ${creatorTable} WHERE id = ?`,
+                    [id])
+            ])
+                .then(([originalLike, creator]) => {
+                    const oldLike = originalLike && originalLike.value || 0
+                    const creatorID = creator.userID
+                    const repChange = creatorID != userID ? (value - oldLike) : 0
+                    db.serialize(function() {
```
NunoDasNeves: I could be wrong, but I'm pretty sure `serialize` does not protect from races between different queries using the same connection - i.e. other node async methods.
It just ensures that one finishes before the next one starts...

NunoDasNeves: I'm pretty sure updating the user table and updating the likes table don't have to be race-free.. but yeah they both have to succeed or fail I guess.

NunoDasNeves: I think don't worry about races rn... it's not worth the effort and won't matter for the demo.

backend/src/models/user.js
NunoDasNeves: I think this is fine... the email and uid may be valid, but this will stop them creating a _profile_ without a displayName

backend/src/models/likes.js
PapaPwn: Doesn't seem to work

backend/src/models/likes.js
PapaPwn: Yeah so its the fact that we guarantee 'BEGIN TRANSACTION' to happen first, but there is still a race cause this isn't all one query.

backend/src/models/likes.js
NunoDasNeves: weird... it works for the SET in course.js

PR #265, opened 2018-10-21
S9 course list filter search

NunoDasNeves:
Search course list to filter
Also, order by name, code, or rating. Rating by default.
Ratings displayed in the form of general review consensus.


NunoDasNeves:
This pr also fixes some bugarinos in table initialization, but now initialization is quite slow; takes 10-15 seconds

PR #266, opened 2018-10-21
S9 my profile improvement

NunoDasNeves:
profile page looking much better yay


NunoDasNeves:
Note this includes al's branch too

NunoDasNeves:
merged into al's branch

PR #269, opened 2018-10-23
Demo data

NunoDasNeves:
Don't merge this!
Contains:
- bugfixes for table init
- reputation is now part of table init
- helper functions for random numbers
- table initialization timers (takes ~60 seconds on my local machine. Longer than before because of bugfixes!)
- linting :p


backend/src/models/db/tables.js
```diff
@@ -211,19 +218,13 @@ function initCourseTable(db) {
 
 function initDegreesTable(db) {
     // Prepare query
-
-    // TODO remove me later
-    const mappedDegrees = degreeData.map(({ name, ...rest }) => ({
-        name: name.startsWith('Bachelor of') ? 'B.' + name.split('Bachelor of')[1] : name,
```
DarkPurple141: Does this get dealt with somewhere else though as these names are really long.

backend/src/models/db/tables.js
```diff
@@ -333,50 +340,52 @@ function initReviewTable(db) {
...
             // Create the question
             const review = {
                 code: code,
-                userID: Math.ceil(Math.random() * NUM_DUMMY_USERS),
-                recommend: Math.round(Math.random()),
-                enjoy: Math.round(Math.random() * (MAX_ENJOY - MIN_ENJOY) + MIN_ENJOY),
-                ...reviewTypes[index] }
+                userID: uid,
+                recommend: getRandomIntInclusive(0,1),
+                enjoy: getRandomIntInclusive(MIN_ENJOY, MAX_ENJOY),
```
PapaPwn: I feel like its odd that we don't have any fake data for the other categories like difficulty and workload. Maybe we should add that in?

backend/src/models/db/tables.js
NunoDasNeves: I'm not sure whatcha mean by 'dealt with'
Maybe we should keep the full name field and also have a shortened name field?

backend/src/models/db/tables.js
NunoDasNeves: i forgot about this! will add now

backend/src/models/db/tables.js
DarkPurple141: Sure. Don't mind. I was thinking we should really remove duplicates like co-op, honours, and remove Bachelor of to be B. and then filter on duplicates.

PR #273, opened 2018-10-26
Demo data fast

NunoDasNeves:
Fails tests because it races tests to initialize database, don't have to time to fix myself.
Otherwise contains changes from day before/day of demo including:
- bugfixes in dummy data init
- couple of small frontend bugfixes


NunoDasNeves:
Does it break what?

NunoDasNeves:
@DarkPurple141 The tests don't wait for the db to be initialized with dummy data, but they rely on the dummy data being there in order to pass.
This used to work because of bugs in the dummy data initialization. In this branch those bugs are fixed, hence the tests are broken too.

PR #274, opened 2018-10-27
S10 UID get change

### LINKED ISSUE #270: Remove Token from GETs
* Pass uid in query params

DarkPurple141:
I don't know if this is better or not. In theory it should be.

Will resolve #270.


DarkPurple141:
Declining this as I don't think it's necessarily better.

PR #279, opened 2018-11-20
S10 sqlerise db init 2

DarkPurple141:


backend/src/models/db/test_constants.js
```diff
@@ -0,0 +1,94 @@
+exports.TEST_DB = './test.db'
+
+exports.NUM_DUMMY_USERS = 50
+
+exports.UNIVERSITY_ID = 1
+
+exports.SAMPLE_QUESTIONS = [
```
DarkPurple141: My only feeling on this was that these questions might be pinned for all courses, or at least a few of them.

backend/src/models/db/init_sql.js
```diff
@@ -151,3 +156,125 @@ function sqlCourse(course) {
...
+            }
+            likes.push(like)
+        }
+    }
+
+    return bulkInsertDB('likes', likes)
+}
+
+/* SQL HELPERS */
+/* THESE ARE ONLY USED FOR TESTING AND ARE VULNERABLE TO SQL INJECTION */
```
DarkPurple141: I don't think we need these at all, it should just work the same way the other sql is generated.

.gitignore
```diff
@@ -24,6 +24,8 @@ yarn-error.log*
 
 backend/public
 backend/db
+backend/src/models/db/sql/test/*
+test.db
```
DarkPurple141: was that not dealt with via the original relative .gitignore?

backend/src/models/db/js/tables.js
```diff
@@ -1,417 +1,3 @@
-const courseData = require('./courses')
```
DarkPurple141: Yay code deletion :)

backend/src/models/db/test_constants.js
DarkPurple141: .. and thus maybe don't belong in test_constants. But whatevs. These in theory will eventually move to prod.

backend/src/models/db/init_sql.js
PapaPwn: So select gets run cause I was too lazy to figure out how to get the returned output from running a shell script, bulkInsert is just a wrapper that returns some SQL (doesn't run it) and runSQL is a wrapper around your method. The reason I chose to uses these is because items with dependencies can be more easily generated, instead of hardcoding queries with assumed ids of things that don't yet exist. This means it should be relatively easy to adapt our testing and get more variety in our test cases. The cost is `~5 seconds`.

backend/src/models/db/test_constants.js
PapaPwn: Oh I see...

.gitignore
PapaPwn: not from what I could tell.

backend/src/models/db/init_sql.js
DarkPurple141: Is the error with travis tests failing related to this though? I'm not not convinced. Understand what you're saying. Also this is for development, so let's not sweat too much ;-/.

.gitignore
DarkPurple141: fair enuf.

PapaPwn: I just realised there was a second .gitignore file. Is this standard practice?

backend/src/models/db/init_sql.js
PapaPwn: Fixed

.gitignore
DarkPurple141: Can be.

PR #293, opened 2018-11-24
Better test data

PapaPwn:
Deterministic testing data with seeded randomness.


frontend/src/store/utils.js
```diff
@@ -27,7 +27,7 @@ export function sortByHotness(list) {
   const d = Date.now()
   ordered.sort(
     ({ likes: l1, timestamp: t1 }, { likes: l2, timestamp: t2 }) =>
-      (l2 + likesInc) / Math.log(d - t2.getTime() + 1000) - (l1 + likesInc) / Math.log(d - t1.getTime() + 1000)
+      (l2 + likesInc) / Math.log(d - t2 + 1000) - (l1 + likesInc) / Math.log(d - t1 + 1000)
```
DarkPurple141: Good catch.

PR #311, opened 2018-12-23
Edit delete

NunoDasNeves:
Just backend so far ...
Happy for someone else to jump on frontend for this if they want to, otherwise I'll keep at it.


NunoDasNeves:
#311

backend/src/controllers/comment.js
```diff
@@ -20,3 +20,17 @@ exports.getComment = function({ params: { cid }, user }, res) {
         res)
         .catch(errorHandler(res))
 }
+
+/* PUT updated comment body */
+exports.putComment = function ({ user, params, body }, res) {
+    body.userID = user.id
+    commentModel.putComment(params.cid, body)
+        .then(() => exports.getComment({ user, params }, res))
```
DarkPurple141: new pattern is not doing this 🥇

DarkPurple141: But I dunno.

backend/tests/routes/review.js
```diff
@@ -122,4 +264,106 @@ describe('Review route testing', function () {
...
+                    location = res.headers.location
+                    deleteRequest = supertest
+                        .delete(location)
+                        .set('Accept', 'application/json')
+                        .set('Authorization', `Bearer ${global.idToken}`)
+                        .expect(204)
+                    return deleteRequest
+                })
+                .then(res => {
+                    getRequest = supertest
```
DarkPurple141: This should be in a new `describe`.

backend/tests/routes/review.js
```diff
@@ -122,4 +264,106 @@ describe('Review route testing', function () {
             )
         })
     })
+
+    describe('PUT /api/course/ACCT1501/review/1/comment/', () => {
+        let postRequest
+        let putRequest
+        let getRequest
+        let location
```
DarkPurple141: again you should probably verify the values of these along the way.

backend/tests/routes/questions.js
```diff
@@ -192,4 +308,106 @@ describe('Test answer routes', () => {
...
+                        .set('Authorization', `Bearer ${global.idToken}`)
+                        .expect(204)
+                    return deleteRequest
+                })
+                .then(res => {
+                    getRequest = supertest
+                        .get(location)
+                        .set('Accept', 'application/json')
+                        .expect('Content-Type', /json/)
+                        // TODO: at the moment there is no 404 implemented for missing posts
```
DarkPurple141: Yeah this is important

backend/src/controllers/comment.js
NunoDasNeves: What is the pattern? As far as i can see the other put routes return the object e.g. putReviewLikes

NunoDasNeves: https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#useful-post-responses

NunoDasNeves: It occurs to me you mean the pattern of calling exports.getThing, rather than the REST pattern of returning the updated resource from a PUT

PR #314, opened 2018-12-29
Error handling 404

NunoDasNeves:
(This isn't finished)

- Cleaner error handling in idiomatic express style (see https://expressjs.com/en/guide/error-handling.html)
- only applied to getCourse() as an example right now
- feedback needed


backend/src/utils/error.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/*
+ * APIError class
+ * Exists so that we can put a status and code in the constructor
+ * Reference:
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
+ */
+exports.APIError = class APIError extends Error {
+    constructor(status = 400, code = 1, ...params) {
```
DarkPurple141: wonder whether params should be named here so it's clearer in the calls.

backend/src/utils/error.js
```diff
@@ -0,0 +1,49 @@
+/*
+ * API error codes
+ * TODO we might want to map error codes to messages or something
```
DarkPurple141: Probs need a error constant file.

backend/src/utils/error.js
NunoDasNeves: The only param we expect to use is 'message' - the first arg of the regular Error constructor. So like this?
`constructor(status = 400, code = 1, message = 'Error', ...params)`

DarkPurple141: Sorry wasn't that clear!

```js
// something like:
function createAPIError({
  status = 400,
  code = 1,
  message = 'Error'
}) {

 return new APIError...
}
```
So the calls are more like:

```js
throw createAPIError({
  code,
  status,
  message,
})
```

backend/src/utils/error.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/*
+ * APIError class
+ * Exists so that we can put a status and code in the constructor
+ * Reference:
+ * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error
+ */
+exports.APIError = class APIError extends Error {
+    constructor({ status = 400, code = 1, message = 'Unkown Error' }) {
```
NunoDasNeves: This is fine right @DarkPurple141 ?

DarkPurple141: Ya.

backend/src/utils/error.js
```diff
@@ -15,35 +18,66 @@
...
+     * Reference for design of these error fields:
+     * https://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#errors
+     * @param {object}
+     *  status: HTTP status code
+     *  code: Error code (see above)
+     *  message: User-readable error message (TODO: different message displayed to user/frontend developer?)
+     *  errors: For validation errors on PUT, PATCH & POST
+    *           list of { code, field, message } for each invalid field:
+     */
+    constructor({ status = 400, code = 1, message = 'Unkown Error', errors = [] }) {
```
DarkPurple141: typo in Unknown

backend/src/utils/error.js
```diff
@@ -15,35 +18,66 @@
...
+ *  @param {object} mapping of SQL error number to HTTP status code
+ *                  supplied error numbers above 50000 are user-defined and
+ *                  should map to API error codes by subtracting 50000
+ */
+exports.translateSQLError = function(obj) {
+    return function (err) {
+        if (err.number && err.number > 50000) {
+            throw new APIError({ status: obj[err.number], code: err.number - 50000, message: err.message })
+        } else {
+            throw err
```
DarkPurple141: comment what case this occurs with

backend/src/auth/index.js
```diff
@@ -19,6 +19,7 @@ module.exports = function (req, _, next) {
         .then(decodedToken => {
             // attach user to request obj
             req.authorized = decodedToken
+            // TODO hmmmm
```
DarkPurple141: ?

backend/src/models/comment.js
```diff
@@ -15,6 +16,19 @@ class Comment {
...
         const [key, value] = Object.entries(queryObject)[0]
+
+        // validation
+        if (!body) {
+            const type = key === 'questionID' ? 'answer' : 'comment'
+            throw new APIError({
+                status: 400,
+                code: 1002,
+                message: `Invalid ${type}`,
+                errors: [{ code: 6002, message: `Invalid ${type} body` }]
```
DarkPurple141: Invalid or just missing?

backend/src/models/question.js
```diff
@@ -18,7 +19,10 @@ class Question {
                 {
                     [QUESTIONS]: { id }
                 })
-            .then(([row]) => row || {})
+            .then(([row]) => {
+                if (row) return row
+                throw new APIError({ status: 404, code: 4001, message: 'The question does not exist' })
```
DarkPurple141: Be good to have less magic numbers. eg what is 4001?

backend/src/utils/error.js
```diff
@@ -3,17 +3,25 @@
...
- *   200: misc subject error
- *   201: subject doesn't exist
- * 3 - course
- *   300: misc course error
- *   301: course doesn't exist
- * 4 - question
- * 5 - review
- * 6 - comment
- * 7 - user/auth
+ * 1000 - misc
```
DarkPurple141: Right.

backend/src/models/question.js
NunoDasNeves: Well it's the error code.
I'm not sure the best way to do this tbh - maybe we should have a constant for each error code? In this case it would be something like `ERR_QUESTION_NOTEXIST` ?

NunoDasNeves:
This is now 'done' barring further discussion/feedback/questions...
I haven't added error handling everywhere yet obviously but the low hanging 404's and general patterns are in there.

Keep in mind it's just backend changes... frontend error handling also needs an overhaul to match.
Pretty sure it doesn't break anything in frontend though tbh haven't tested super extensively...

backend/src/models/comment.js
```diff
@@ -63,7 +63,7 @@ class Comment {
             {
                 [COMMENTS]: { [key]: value }
             })
-            // TODO: check comment exists and 404 if not
+            // TODO: check parent object exists and 404 if not
```
DarkPurple141: 400 if not ?

backend/src/models/review.js
```diff
@@ -75,25 +83,41 @@ class Review {
...
-        if (enjoy < MIN_ENJOY || enjoy > MAX_ENJOY) throw Error('Invalid enjoy value')
-
-        ;[difficulty, teaching, workload].forEach(item => {
-            if (item < MIN_OPTION || item > MAX_OPTION) throw Error('Invalid difficulty, teaching or workload value')
-        })
+        // validation
+        const errors = []
+        if (!title) errors.push({ code: 5002, message: 'Review must have a title' })
+        if (!body) errors.push({ code: 5003, message: 'Review must have a body' })
+        if (recommend !== DONT_RECOMMEND && recommend !== RECOMMEND) {
```
DarkPurple141:
```js
if (!(recommend in [DONT_RECOMMEND, RECOMMEND]))
``` 
~~maybe~~. No, my bad.

backend/src/models/comment.js
NunoDasNeves: I'm not sure - I thought 404 because you're trying to (e.g.) POST an answer to a question that doesn't exist.

PR #318, opened 2019-01-11
Significant Improvement to Dev Database through Docker

DarkPurple141:
Blazing fast initialisation speeds, that don't rely on Azure box.

Will look at this on weekend:
https://github.com/twright-msft/mssql-node-docker-demo-app

Seems like initialisation not done when tests run.


NunoDasNeves:
> > Do you need to create the database in the docker script?
> > In mine, after creating the container I do:
> > ```
> > sudo docker exec -it <container-name> /opt/mssql-tools/bin/sqlcmd \
> >    -S localhost -U SA -P <password> \
> >    -Q 'CREATE DATABASE <dbname>'
> > ```
> 
> Interestingly I didn't need this, but I'll add anyway.

Oh that's odd... it definitely doesn't work for me without that - there doesn't seem to be any `CREATE DATABASE` statement in the repo otherwise.

PR #322, opened 2019-01-17
Sessions Frontend

DarkPurple141:
Closeeerrr.

### DIFF
--- a/docker.sh
+++ b/docker.sh
@@ -1,9 +1,16 @@
-#!/bin/bash -e
+#!/bin/bash
+
+DOCKER_NAME='smartcourse-testing'
+
+docker kill $DOCKER_NAME
+docker rm $DOCKER_NAME
+
+set -ex
 
 docker pull microsoft/mssql-server-linux:2017-latest
-# docker build --tag 'smartcourse-testing' .
+# docker build --tag $DOCKER_NAME .
 docker run \
-    --name 'smartcourse-testing' \
+    --name $DOCKER_NAME \
     --env 'ACCEPT_EULA=Y' \
     --env "MSSQL_SA_PASSWORD=$AZURE_SQL_PASSWORD" \
     -p 1433:1433 \
@@ -11,7 +18,7 @@ docker run \
 
 sleep 20
 
-docker exec -it 'smartcourse-testing' /opt/mssql-tools/bin/sqlcmd \
+docker exec -it $DOCKER_NAME /opt/mssql-tools/bin/sqlcmd \
    -S localhost,1433 -U SA -P "$AZURE_SQL_PASSWORD" \
    -Q 'CREATE DATABASE testdb'
  
--- a/docker.sh
+++ b/docker.sh
@@ -7,7 +7,6 @@ docker rm $DOCKER_NAME
 
 set -ex
 
-docker pull microsoft/mssql-server-linux:2017-latest
 # docker build --tag $DOCKER_NAME .
 docker run \
     --name $DOCKER_NAME \
--- a/frontend/src/components/Course/FeedCard.vue
+++ b/frontend/src/components/Course/FeedCard.vue
@@ -50,7 +50,9 @@ export default {
     id: String,
     cardType: String,
     recommend: Boolean,
-    teachingPeriod: { type: String, default: '18s2' }
+    // it's an id not a string, can be used
+    // to index the sessions array
+    session: { type: Number, default: 0 }
   },
   components: {
     Card,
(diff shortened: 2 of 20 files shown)


docker.sh
```diff
@@ -1,17 +1,24 @@
...
+#!/bin/bash
+
+DOCKER_NAME='smartcourse-testing'
+
+docker kill $DOCKER_NAME
+docker rm $DOCKER_NAME
+
+set -ex
 
 docker pull microsoft/mssql-server-linux:2017-latest
```
NunoDasNeves: Ummm i'm not sure if this is needed for travis or w/e, but if we're running this script locally we don't want it re-downloading the image every time, so maybe it should check the output of `docker image ls | grep mssql | grep 2017-latest` before downloading?

frontend/src/components/Reviews/ReviewCard.vue
```diff
@@ -28,7 +32,17 @@ export default {
...
+    recommend: Boolean,
+    session: {
+      default: 0,
+      type: Number
+    }
+  },
+  computed: {
+    sessionShortName() {
+      return this.$store.getters.sessions.length &&
+        this.$store.getters.sessions[this.session].shortName
```
NunoDasNeves: This displays the wrong session because database id's start at 1 and array indicies start at 0

frontend/src/components/Reviews/ReviewForm.vue
```diff
@@ -9,6 +9,9 @@
       <span class="required">*</span>
       <textarea placeholder="Your review here..." v-model="body"></textarea><br>
 
+      <OptionHeader>What session did you take the course?</OptionHeader>
+      <Selector :items="sessions" v-model="session" :placeholder="'Select a session..'"/>
```
NunoDasNeves: This needs some improvement imo;
- Should show in reverse order (later sessions at the top)
- Should only show up to current session
- Searchable maybe?

NunoDasNeves:
One weird thing seems to be that building frontend and viewing `localhost:3000` in the browser doesn't seem to work on this branch... 8080 works though
screenshot from 2019-01-18 11-20-12

docker.sh
DarkPurple141: Was slightly lazy; I need @PapaPwn to write it better. It's a simple if statement.

DarkPurple141: Also I think if docker is running it cache's the image.

NunoDasNeves:
> > One weird thing seems to be that building frontend and viewing `localhost:3000` in the browser doesn't seem to work on this branch... 8080 works though.
> > Console says this:
> > screenshot from 2019-01-18 11-20-12
> 
> Did you do a fresh npm i?

I just did, annnnd still broken :(

It looks like the backend is serving index.html when app.js is requested, so it thinks the html is js

NunoDasNeves:
Also I get the `TypeError: Property 'handleEvent' is not callable` with 8080 also..

frontend/src/components/Reviews/ReviewForm.vue
NunoDasNeves: Dropdown is good. I think don't worry about searching for now; with the other improvements it'll be useable

docker.sh
NunoDasNeves: Here:
```if sudo docker image ls | grep mssql | grep -q 2017-latest; then
  docker pull microsoft/mssql-server-linux:2017-latest
fi```
Actually i'll just test this and push it to this branch

Comment (NunoDasNeves) [reply docker.sh]: So I think you can just delete this line @DarkPurple141

File: frontend/src/components/Course/FeedCard.vue [hunk 4]
```diff
@@ -66,6 +68,10 @@ export default {
     },
     positiveOrNegativeText() {
       return this.recommend ? 'Recommended' : 'Not Recommended'
+    },
+    teachingPeriod() {
+      return this.$store.getters.sessions.length &&
+        this.$store.getters.sessions[this.session].shortName
```
NunoDasNeves: this needs the -1 as well probably

DarkPurple141: Ahh yep.

DarkPurple141:
> See comment.
> Also, review view is broken.
> Also, look at `http://localhost:3000/api/uni/sessions` - the data is duplicated many times - something is messed up in the initialization

The duplication was because on a fresh db; it was initialising twice, but this was a hack because I wanted to make sure sessions got populated on staging. Have since removed it. If you delete ur db it'll die. Review view is broken eh.

NunoDasNeves:
I think this is ok... though i think maybe you should still add the current session endpoint etc before merging. or just hardcode it in frontend for now

Issue #324, opened 2019-01-20
Frontend: Provide prompt for no Reviews, no Questions under tab view


DarkPurple141:
<img width="814" alt="screen shot 2019-01-20 at 12 02 16 pm" src="https://user-images.githubusercontent.com/13936670/51433974-d3ce2a80-1cab-11e9-915b-86e8d189d8ce.png">

DarkPurple141:
Resolved by #326

Issue #275, opened 2018-10-31
Pinned Questions for all courses


DarkPurple141:
As a User I should be greeted with selected questions that are likely to be more common, and thus should be pinned. These should be generated and created by the admin user.

DarkPurple141:
This is mostly done; just need the FAQ part to be reflected in frontend.

DarkPurple141:
Resolved in 326.

PR #326, opened 2019-01-20
Pinned Qs, 

DarkPurple141:
Added support for the pinned badge: called 'FAQ', added a placeholder for when there are no reviews.

Resolves requirements for #324, #275.

Just need to re-add FAQ questions to production build.


DarkPurple141:
Already in the response. Has to be set on init though.

PR #328, opened 2019-01-22
Basic moderation backend

NunoDasNeves:
fixed some test stuff
moved backend test secrets out of repo
superuser now exists, can delete and edit arbitrary comments/questions/reviews

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -6,7 +6,7 @@
     "start": "cross-env NODE_ENV=test LOG=true npx nodemon ./src/bin/www",
     "lint": "npx eslint src/**/*.js tests/**/*.js --fix",
     "test:lint": "npx eslint src/**/*.js tests/**/*.js",
-    "test:src": "cross-env NODE_ENV=test npx nyc mocha --file tests/utils/*.js --file tests/routes/*.js --timeout 75000",
+    "test:src": "cross-env NODE_ENV=test npx nyc mocha --file tests/utils/*.js --file tests/routes/*.js --timeout 10000",
     "test": "npm run test:lint && npm run test:src",
     "coverage": "npx nyc report --reporter=html --reporter=text-lcov | coveralls"
   },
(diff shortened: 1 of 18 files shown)


backend/package.json
```diff
@@ -6,7 +6,7 @@
     "start": "cross-env NODE_ENV=test LOG=true npx nodemon ./src/bin/www",
     "lint": "npx eslint src/**/*.js tests/**/*.js --fix",
     "test:lint": "npx eslint src/**/*.js tests/**/*.js",
-    "test:src": "cross-env NODE_ENV=test npx nyc mocha tests/**/*.js --timeout 75000",
+    "test:src": "cross-env NODE_ENV=test npx nyc mocha --file tests/utils/*.js --file tests/routes/*.js --timeout 75000",
```
DarkPurple141: This timeout is probably excessive now.

backend/src/models/constants.js
```diff
@@ -16,6 +16,12 @@ exports.MAX_ENJOY = 5
 exports.MIN_OPTION = 0
 exports.MAX_OPTION = 3
 
+// Permissions Constants (TODO:this properly, more granularity etc)
```
DarkPurple141: Although verbose; I think `PERMISSIONS_*` is better.

backend/src/models/constants.js
```diff
@@ -356,6 +362,10 @@ exports.TABLE_COLUMNS = {
         picture: {
             type: TYPES.VarChar,
             options: { nullable: true }
+        },
+        permissions: {
```
DarkPurple141: This should probably be a table? To enforce `enum`-ness.

NunoDasNeves: Yeah maybe but then joinszzz hnng. We have constants - are they not good enough?
We will need another table for a more complex permissions model but until then...cbf?

DarkPurple141: okeyyy

PR #336, opened 2019-01-30
Admin user creation

NunoDasNeves:
Create user profiles in the backend for each superuser (Nuno, Alex, Papa).

### DIFF
--- a/backend/src/models/db/index.js
+++ b/backend/src/models/db/index.js
@@ -13,6 +13,7 @@ const {
     sqlComments,
     sqlSessions,
     sqlLikes,
+    sqlAdminUsers,
     sqlUsers,
     unswDataInitialised,
     testDataInitialised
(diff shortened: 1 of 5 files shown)


backend/src/models/db/index.js
```diff
@@ -65,11 +66,12 @@ class DB {
...
+                        await sqlAdminUsers(connection)
+                        await sqlQuestions(connection)
                     }
 
                     // Initialise test databases
                     if (!PRODUCTION && !await testDataInitialised(connection)) {
-                        await sqlQuestions(connection)
                         await sqlReviews(connection)
                         await sqlComments(connection)
                         await sqlLikes(connection)
```
DarkPurple141: Maybe add sqlUsers here for non-Admin

NunoDasNeves: it's on the next line ^_^

PR #339, opened 2019-02-05
Separate db init from backend init

NunoDasNeves:
Scripts directory moved into backend to prep for splitting backend and frontend repos
`backend/scripts/db` contains database initialization. It's basically just some of the code that used to be in `backend/src/models/db`
Tests should work, local dev should work the same as before
Sensible defaults depending on `NODE_ENV`, but you can use the script to do other things like clear production or only initialize with static data for example.
Prompts twice for confirmation when doing anything potentially unsafe on production.

`NODE_ENV` can be `test`, `production` or `staging`. Anyplace where it used to be `development` should be `staging` - I'm not sure what it is on Azure staging server rn (how to change dis @PapaPwn ?)


backend/package.json
```diff
@@ -3,12 +3,13 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "cross-env NODE_ENV=test LOG=true npx nodemon ./src/bin/www",
+    "start": "npm run db-init && cross-env NODE_ENV=test LOG=true npx nodemon ./src/bin/www",
     "lint": "npx eslint src/**/*.js tests/**/*.js --fix",
     "test:lint": "npx eslint src/**/*.js tests/**/*.js --fix",
-    "test:src": "cross-env NODE_ENV=test npx nyc mocha --file tests/utils/*.js --file tests/routes/*.js --timeout 10000",
+    "test:src": "npm run db-init && cross-env NODE_ENV=test npx nyc mocha --file tests/utils/*.js --file tests/routes/*.js --timeout 10000",
```
DarkPurple141: curious -- is cross-env not persisting across this and the above call?

backend/scripts/db/config.js
```diff
@@ -0,0 +1,77 @@
+const stdio = require('stdio');
```
DarkPurple141: semi

backend/scripts/db/config.js
```diff
@@ -0,0 +1,77 @@
...
+    },
+    PRODUCTION: {
+        drop: 'none',
+        create: 'all',
+        init: 'none'
+    }
+}
+
+const STDIO_OPTIONS = {
+    'drop': {args: 1, description: 'Drop tables first - \'all\', \'data\' or \'none\'. \'data\' includes users, reviews, questions, comments, likes, reports etc..'},
```
DarkPurple141: quoted keys

backend/scripts/db/config.js
```diff
@@ -0,0 +1,77 @@
...
+            return acc
+        }, [])
+
+    if (invalid.length) {
+        console.warn(`Invalid argument value for these options: ${invalid}`)
+        process.exit(1)
+    }
+
+    // default to most conservative config
+    const default_config = TESTING ? DEFAULT_CONFIG.TESTING : (STAGING ? DEFAULT_CONFIG.STAGING : DEFAULT_CONFIG.PRODUCTION)
```
DarkPurple141: If you're going to next ternary, do it over two lines

backend/scripts/db/index.js
```diff
@@ -0,0 +1,62 @@
...
+
+        if (naughty.length) {
+            // print the messages
+            naughty.forEach(message => { console.log(message) })
+            // prompt twice for extra safety!
+            const proceed = await new Promise((resolve, reject) => {
+                stdio.question('continue? Y/n', (err, text) => {
+                    if (err) {
+                        reject(err)
+                        return
```
DarkPurple141: Do all these returns do anything?

backend/src/index.js
```diff
@@ -71,8 +71,9 @@ app.use(APIErrorHandler)
 /*
  * Connect to SQL server
  */
-db.init()
-    .then(() => console.log('App ready!'))
-    .then(() => app.emit('ready'))
+db.on('ready', () => {
```
DarkPurple141: where does this ready event come from?

backend/src/models/db/index.js
```diff
@@ -39,58 +26,15 @@ class DB {
...
-                    // Log completion time
-                    timeList.push(Date.now() / 1000)
-                    console.log(`Done creating database! (${((timeList[1] - timeList[0])).toFixed(3)})`)
-
-                    // Done
-                    resolve()
-                })
-                connection.execSql(request)
+                // emit ready event once only when a connection becomes available
+                if (!this.ready) {
```
DarkPurple141: does `this.ready` need to exist? Assume it's used elsewhere

DarkPurple141: Can't see where this used unless just to be stateful

backend/src/models/db/index.js
```diff
@@ -138,4 +82,6 @@ class DB {
     }
 }
 
+//util.inherits(DB, EventEmitter)
```
DarkPurple141: delete>

backend/scripts/db/index.js
NunoDasNeves: kinda, but not really. fixed now

backend/src/index.js
NunoDasNeves: the database emits ready when the DB class's constructor successfully connects to the database

backend/src/models/db/index.js
NunoDasNeves: it's used to ensure the ready event is only fired once (because a bunch of connections actually get created)
it's not used elsewhere

backend/scripts/db/config.js
NunoDasNeves: que?

NunoDasNeves: oooh

backend/package.json
NunoDasNeves: i'm not sure - didn't try it without..
looking at the cross-env docs it seems like you need to use `cross-env-shell` to have environment variables persist through an entire shell command ... maybe?

PR #340, opened 2019-02-07
authCV initialized to null, getter used in checkAuth

NunoDasNeves:
idk how to enable prerender on staging - it seems like it's enabled everywhere according to `src/backend/index.js` ??

### DIFF
--- a/frontend/.env.production
+++ b/frontend/.env.production
@@ -1,4 +1,3 @@
 NODE_ENV=production
 VUE_APP_TITLE=SmartCourse
 VUE_APP_API_URL=https://smartcourse.me/api
-VUE_APP_URL=https://smartcourse-staging.azurewebsites.net
--- a/frontend/.env.production
+++ b/frontend/.env.production
@@ -1,3 +1,4 @@
 NODE_ENV=production
 VUE_APP_TITLE=SmartCourse
 VUE_APP_API_URL=https://smartcourse.me/api
+VUE_APP_URL=https://smartcourse-staging.azurewebsites.net
--- a/frontend/.env.staging
+++ b/frontend/.env.staging
@@ -1,3 +1,4 @@
 NODE_ENV=staging
 VUE_APP_TITLE=SmartCourse (staging)
 VUE_APP_API_URL=https://smartcourse-staging.azurewebsites.net/api
+VUE_APP_URL=https://smartcourse.me
--- a/frontend/.env.staging
+++ b/frontend/.env.staging
@@ -1,4 +1,4 @@
 NODE_ENV=staging
 VUE_APP_TITLE=SmartCourse (staging)
 VUE_APP_API_URL=https://smartcourse-staging.azurewebsites.net/api
-VUE_APP_URL=https://smartcourse.me
+VUE_APP_URL=https://smartcourse-staging.azurewebsites.net
--- a/frontend/src/store/auth/index.js
+++ b/frontend/src/store/auth/index.js
@@ -173,7 +173,7 @@ const actions = {
    * Logs into firebase and retrieves the profile
    * If anything fails it clears everything
    */
-  async checkAuth({ commit, dispatch, state, getters }) {
+  async checkAuth({ commit, dispatch, state }) {
     commit('SET_LOADING', true)
     try {
       // returns user object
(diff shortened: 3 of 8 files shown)


frontend/src/store/auth/index.js
```diff
@@ -27,7 +27,10 @@ const getters = {
   userAuthObject: ({ userAuthObject }) => userAuthObject,
   loading: ({ loading }) => loading,
   error: ({ error }) => error,
-  authCV: ({ authCV }) => authCV
+  authCV: ({ authCV }) => {
```
DarkPurple141: This doesn't work. It'll just alter the param, not the underlying CV.

I suggest this:

```js
const _authCV = new CV()

const state = {...}

const getters = { authCV: () => _authCV }
const mutations = { ...SIGNAL_AUTH_CV() { _authCV.signal() }}
```

frontend/src/store/auth/index.js
```diff
@@ -12,7 +12,7 @@ const state = {
   // our own user data
   profile: null,
   // condition variable for app to wait on while waiting for auth to resolve on boot
-  authCV: new CV()
+  authCV: null
```
DarkPurple141: and then just remove this.

frontend/src/store/auth/index.js
```diff
@@ -193,7 +193,7 @@ const actions = {
 
     // signal the CV so the app can continue loading and use the JWT token in its requests
     // Note we _need_ to do this before returning!
-    commit('SIGNAL_AUTH_CV')
```
DarkPurple141: And then retain this as a commit

frontend/src/store/auth/index.js
NunoDasNeves: can't i just not destructure state? that seems to work
```
  authCV: (state) => {
    if (state.authCV === null) state.authCV = new CV()
    return state.authCV
  }
```

NunoDasNeves: Also this is why we should be using Rust for our frontend ^_^

NunoDasNeves:
> I mean I guess. Seems like violation of getters being non mutating. Also not sure if authCV is really application state.

hok then

DarkPurple141:
@NunoDasNeves also the previous deploy was broken so 👍

frontend/vue.config.js
```diff
@@ -15,10 +17,26 @@ const productionPlugins = [
       // so we create a snapshot of the page after the
       // `data-view` attribute exists in the DOM.
       renderAfterElementExists: '[data-view]'
-    })
+    }),
+    postProcess (renderedRoute) {
+      // Ignore any redirects.
+      renderedRoute.route = renderedRoute.originalRoute
+      // Basic whitespace removal. (Don't use this in production.)
```
NunoDasNeves: don't use in production? o.O

frontend/vue.config.js
```diff
@@ -15,10 +17,26 @@ const productionPlugins = [
...
+      renderedRoute.html = renderedRoute.html.split(/<div id="app"/)
```
NunoDasNeves: this has been added to index.html already

DarkPurple141: Oh this is a quote I copy-pasta'd. Just wanted to see if this works to stop re-render.

frontend/.env.production
```diff
@@ -1,3 +1,4 @@
 NODE_ENV=production
 VUE_APP_TITLE=SmartCourse
-VUE_APP_API_URL=https://smartcourse.me/api
\ No newline at end of file
+VUE_APP_API_URL=https://smartcourse.me/api
+VUE_APP_URL=https://smartcourse-staging.azurewebsites.net
```
NunoDasNeves: uhh is this right?

frontend/.env.staging
```diff
@@ -1,3 +1,4 @@
 NODE_ENV=staging
 VUE_APP_TITLE=SmartCourse (staging)
 VUE_APP_API_URL=https://smartcourse-staging.azurewebsites.net/api
+VUE_APP_URL=https://smartcourse.me
```
NunoDasNeves: i think you got these switcharoo'd

PR #341, opened 2019-02-07
Email verification

NunoDasNeves:
it works
also updated password reset to have a 'continue' button that links back to the site from firebase

### DIFF
--- a/frontend/src/router/routes.js
+++ b/frontend/src/router/routes.js
@@ -117,6 +117,11 @@ export default new Router({
       component: () => import('../views/SignUp')
     },
     {
+      path: '/verify-email',
+      name: 'Verify Email',
+      component: () => import('../views/VerifyEmail')
+    },
+    {
       path: '/create-profile',
       name: 'Create Profile',
       component: () => import('../views/CreateProfile')
--- a/frontend/src/store/auth/index.js
+++ b/frontend/src/store/auth/index.js
@@ -4,10 +4,8 @@ import CV from './CV'
 
 import { createProfile, updateProfile, getSelf } from '@/utils/api/auth'
 
-// TODO: more idiomatic way to do this? Router something?
-const APP_URL = `${window.location.protocol}//${window.location.host}`
-const CONTINUE_VERIFY_URL = `${APP_URL}/create-profile`
-const CONTINUE_RESET_URL = `${APP_URL}/login`
+const CONTINUE_VERIFY_URL = `${process.env.VUE_APP_URL}/create-profile`
+const CONTINUE_RESET_URL = `${process.env.VUE_APP_URL}/login`
 
 // condition variable for app to wait on while waiting for auth to resolve on boot
 const _authCV = new CV()
--- a/frontend/src/store/auth/index.js
+++ b/frontend/src/store/auth/index.js
@@ -5,9 +5,9 @@ import CV from './CV'
 import { createProfile, updateProfile, getSelf } from '@/utils/api/auth'
 
 // TODO: more idiomatic way to do this? Router something?
-const _baseURL = `${window.location.protocol}//${window.location.host}`
-const _continueVerifyURL = `${_baseURL}/create-profile`
-const _continueResetURL = `${_baseURL}/login`
+const APP_URL = `${window.location.protocol}//${window.location.host}`
+const CONTINUE_VERIFY_URL = `${APP_URL}/create-profile`
+const CONTINUE_RESET_URL = `${APP_URL}/login`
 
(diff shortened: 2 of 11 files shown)


NunoDasNeves:
tests fail cos travis needs updating with new backend users (that all have their emails verified)

NunoDasNeves:
Also, this is a bit jank but until admin (@NunoDasNeves , @DarkPurple141 , @PapaPwn ) users verify their emails they won't be able to edit their profile after this PR, cos of the middleware stacking in backend.
So just verify your email if this is an issue

frontend/src/router/routes.js
```diff
@@ -116,6 +116,11 @@ export default new Router({
       name: 'Sign Up',
       component: () => import('../views/SignUp')
     },
+    {
+      path: '/verify-email',
```
DarkPurple141: new route eh?

frontend/src/store/auth/index.js
```diff
@@ -4,6 +4,11 @@ import CV from './CV'
 
 import { createProfile, updateProfile, getSelf } from '@/utils/api/auth'
 
+// TODO: more idiomatic way to do this? Router something?
+const _baseURL = `${window.location.protocol}//${window.location.host}`
```
DarkPurple141: base url should be exposed by as VUE_APP_URL

frontend/src/store/auth/index.js
```diff
@@ -18,10 +23,12 @@ const state = {
...
 // logged into firebase (authenticated account)
   isFirebaseAuthorised: ({ userAuthObject }) => !!userAuthObject,
+
+  emailVerified: ({ userAuthObject }) => !!userAuthObject && userAuthObject.emailVerified,
   // TODO not sure if this is useful...
   hasProfile: ({ profile }) => !!profile,
-  // logged into backend (existing profile) and authed with firebase
-  isLoggedIn: ({ profile, userAuthObject }) => !!profile && !!userAuthObject,
+  // logged into backend (existing profile) and authed with firebase with a verified email address
+  isLoggedIn: ({ profile, userAuthObject }) => !!profile && !!userAuthObject && userAuthObject.emailVerified,
```
DarkPurple141: have noticed recently that:

```js
!!{} = true
```

DarkPurple141:
> Also, this is a bit jank but until admin (@NunoDasNeves , @DarkPurple141 , @PapaPwn ) users verify their emails they won't be able to edit their profile (or use some other user-related endpoints) after this PR, cos of the middleware stacking in backend.
> So just verify your email if this is an issue..

I haven't recevied an email? Do I need to login?

frontend/src/store/auth/index.js
NunoDasNeves: Yeah... That shouldn't break stuff here. But maybe it would be better to change these to `thing === null` to be more explicit

frontend/src/router/routes.js
NunoDasNeves: Yeah honestly the whole signup flow could probably be in one page tbh. But meh

NunoDasNeves:
Err good point I'll just send out the verification emails I guess

frontend/src/store/auth/index.js
NunoDasNeves: hmmm - this won't work on localhost however - I'm not sure why but you removed `VUE_APP_URL=localhost:8080` from the .env file...

In any case the port may change (I use syncthing and it uses port 8080 by default :/)

So either this should stay or VUE_APP_URL should be fixed, but idk how to make it work for every environment

NunoDasNeves: i think imma leave it for now

frontend/src/store/auth/index.js
```diff
@@ -5,9 +5,9 @@ import CV from './CV'
 import { createProfile, updateProfile, getSelf } from '@/utils/api/auth'
 
 // TODO: more idiomatic way to do this? Router something?
-const _baseURL = `${window.location.protocol}//${window.location.host}`
-const _continueVerifyURL = `${_baseURL}/create-profile`
-const _continueResetURL = `${_baseURL}/login`
+const APP_URL = `${window.location.protocol}//${window.location.host}`
+const CONTINUE_VERIFY_URL = `${APP_URL}/create-profile`
```
DarkPurple141: VUE_APP_URL? Why not?

NunoDasNeves: screenshot from 2019-02-08 15-08-36
screenshot from 2019-02-08 15-08-53

DarkPurple141: This is a meme

PR #343, opened 2019-02-09
Fixed some broken things in profiles

DarkPurple141:
* Profile fixes
* Amend user mapper
* Redundant styles
* aria-attributes

### DIFF
--- a/frontend/src/utils/api/user.js
+++ b/frontend/src/utils/api/user.js
@@ -14,8 +14,12 @@ export function userMapper({
   joined,
   ...rest
 }) {
+  if (!joined) {
+    return
+  }
+
   return {
     ...rest,
-    joined: format(joined * 1000, 'dd/MM/yyyy')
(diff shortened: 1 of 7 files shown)


frontend/src/utils/api/user.js
```diff
@@ -14,8 +14,12 @@ export function userMapper({
   joined,
   ...rest
 }) {
+  if (!joined) {
+    return
```
NunoDasNeves: return nothing?

NunoDasNeves: shouldn't it return `rest` or `{ ...rest }` ?

NunoDasNeves: just to stop errors happening idk

DarkPurple141: Hmm. Yeah I dunno. To me it implies the user doesn't exist, and other fields will also error.

DarkPurple141: `joined` is compulsory I'm pretty sure.

PR #345, opened 2019-02-09
Bugfix: Ratings fixes

DarkPurple141:


frontend/src/views/course/index.vue
```diff
@@ -71,19 +71,18 @@ export default {
...
   },
   beforeRouteUpdate ({ params: { code } }, from, next) {
-    // called when the route that renders this component has changed,
-    // but this component is reused in the new route.
-    // For example, for a route with dynamic params `/foo/:code`, when we
-    // navigate between `/foo/1` and `/foo/2`, the same `Foo` component instance
-    // will be reused, and this hook will be called when that happens.
-    // has access to `this` component instance.
-    if (this.code && this.code !== code) {
+    // BUG as far as I'm aware this is only called
```
NunoDasNeves: i think this fixed a bug where data was sticking between coursesss... but if that's happening again...

DarkPurple141: Doesn’t seem to be doing anything.

DarkPurple141: The data isn’t sticking anyway; it’s the api that’s not wkrking

PR #355, opened 2019-02-11
One review per user per course

### LINKED ISSUE #348: User can write multiple reviews of same Course
## Reproduce steps
1. Write a review
2. Write another review.

NunoDasNeves:
resolves #348 
An extra SQL query on getCourse to determine if a user has reviewed the course (this could potentially just be done in the getCourse SQL however...), which is returned in the course object to the frontend.

### DIFF
--- a/backend/src/controllers/course.js
+++ b/backend/src/controllers/course.js
@@ -14,8 +14,10 @@ exports.getCourses = function (_, res, next) {
 }
 
 /* Get specifc course data */
-exports.getCourse = function ({ params }, res, next) {
-    courseModel.getCourse(params.code)
+exports.getCourse = function ({ params, user }, res, next) {
+    const userID = (user && user.id) || ANONYMOUS
+    Promise.all([courseModel.getCourse(params.code), reviewModel.userHasReviewed(params.code, userID)])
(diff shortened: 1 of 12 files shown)


backend/src/controllers/course.js
```diff
@@ -14,8 +14,10 @@ exports.getCourses = function (_, res, next) {
 }
 
 /* Get specifc course data */
-exports.getCourse = function ({ params }, res, next) {
-    courseModel.getCourse(params.code)
+exports.getCourse = function ({ params, user }, res, next) {
+    const userID = (user && user.id) || ANONYMOUS
+    Promise.all([courseModel.getCourse(params.code), reviewModel.userHasReviewed(params.code, userID)])
```
DarkPurple141: Hmm danger here is I can still navigate directly to the `course/review/new` route right?

NunoDasNeves: Yes that's true, although i think it's an unlikely path.
Note that the user will still be unable to POST a new review cos it's enforced on the postReview route.
I did notice that if an error occurs on posting a review, it still tries to navigate to the new review.
Might just push a fix for that and be done with it for now.

A more foolproof fix would need to hit the course endpoint (or dedicated 'hasReviewed' endpoint?) in NewReview.

PR #356, opened 2019-02-11
Do auth while loading

NunoDasNeves:
I noticed app load still looks very ugly due to checkAuth blocking the router and therefore stopping the main routerview showing anything, and the buttons in the nav changing while the page is loading.
So I added startLoad and endLoad to it and now it's all hidden! yay

### DIFF
--- a/frontend/src/store/auth/index.js
+++ b/frontend/src/store/auth/index.js
@@ -1,6 +1,7 @@
 // firebase authentication class
 import auth from './config'
 import CV from './CV'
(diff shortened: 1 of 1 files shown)


NunoDasNeves:
`#shell` just (visually) covers `#app` while loading right?
In this case we do still need the authCV, cos it's purpose is actually to block further requests from going out until checkAuth resolves (specifically, GET requests that return stuff dependent on the auth token)

PR #357, opened 2019-02-11
Mobile Menu

DarkPurple141:
After reverting previous direct dev commits.

screen shot 2019-02-11 at 7 46 26 pm

### DIFF
--- a/frontend/src/components/Nav/MiniMenu.vue
+++ b/frontend/src/components/Nav/MiniMenu.vue
@@ -1,0 +1,85 @@
+<template>
+  <div class="mini-menu">
+    <i class="material-icons">{{ toggled ? 'close' : 'menu'}}</i>
+    <div :class="['mini-menu__items', toggled ? 'toggled' : 'closed' ]">
+      <slot/>
+      <router-link v-for="item in items"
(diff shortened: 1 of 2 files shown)


NunoDasNeves:
Hmm, it looks like the logout button isn't quite right on my Firefox devtools phone-thingy:
screenshot from 2019-02-11 20-07-21

DarkPurple141:
> Hmm, it looks like the logout button isn't quite right on my Firefox devtools phone-thingy:
> screenshot from 2019-02-11 20-07-21

Hmm. Sec

frontend/src/components/Nav/MiniMenu.vue
```diff
@@ -20,15 +20,23 @@
           class="menu-item"
         >{{ item.text }}</router-link>
       </div>
-      <slot/>
+      <div
```
NunoDasNeves: Looks great but shouldn't it be a button or h3 or something with `cursor: pointer;` ? I guess this doesn't matter much given it's for mobile

DarkPurple141: yeah because it's mobile doesn't matter.

PR #359, opened 2019-02-12
Homepage update

NunoDasNeves:
just a small update to simplify the copy on the home page for launching

### DIFF
--- a/frontend/src/components/Nav/MiniMenu.vue
+++ b/frontend/src/components/Nav/MiniMenu.vue
@@ -127,7 +127,7 @@ h2 {
   right: -20px;
   height: calc(100vh - 64px);
   overflow: hidden;
-  z-index: 98;
+  z-index: var(--z-index-mini-menu);
 }
 
 .mini-menu__items--logout {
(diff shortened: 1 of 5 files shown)


NunoDasNeves:
screenshot from 2019-02-12 12-46-21

frontend/src/components/Nav/MiniMenu.vue
```diff
@@ -127,7 +127,7 @@ h2 {
   right: -20px;
   height: calc(100vh - 64px);
   overflow: hidden;
-  z-index: 99;
+  z-index: 98;
```
DarkPurple141: can u make this a css var in the root var declarations and then apply the same to both so we don't accidentally do this again.

this would then become:

```css
z-index: var(--loading-overlay-z-index) - 1;
```

DarkPurple141:
Also, on mobile; the boxes are now quite narrow.

frontend/src/components/Nav/MiniMenu.vue
NunoDasNeves: I think I'll give each use of z-index a variable. Then they can all be changed in once place!

PR #364, opened 2019-02-13
Hompage / drawer tweaks

### LINKED ISSUE #363: Fix: Minor design tweaks
* Tweak Homepage Designs
* Tweak Drawer title sizes

DarkPurple141:
Resolves #363.

image

Interestingly @NunoDasNeves -> maybe the stripped down style makes more sense? I discovered it by accident when I was component-ising things. Also ignore the alignments and stuff.


NunoDasNeves:
Hmmmm. I kinda like it, but I think the existing is fine. 
Don't know how much I like the tiny icons, but I think if they were big then it'd be too monochrome

DarkPurple141:
Just for reference outside of Slack. I think this is the way to go tbh. And then next high prio is feed. I've created an API issue as well.

PR #370, opened 2019-02-23
404 bugfix

### LINKED ISSUE #325: Frontend: 404 Bug
Reviews/Questions still load for pages with no-id.

eg.

screen shot 2019-01-20 at 12 09 40 pm

NunoDasNeves:
resolves #325 

Note that this does NOT fix the fact that you can still view any question/review with any course in the url.
e.g. you can view `/course/ACCT1501/question/50` of the test data even though question 50 is for a different course. This will be fixed in backend

### DIFF
--- a/src/store/utils.js
+++ b/src/store/utils.js
@@ -1,16 +1,21 @@
 
+import APIError from '../utils/api/errors'
+
 // factories
 export function doRequestFactory(REQUEST, COMMITS) {
   return async function ({ commit }, { action, args, load = 'TOGGLE_LOADING' }) {
     if (load) commit(load, true)
+    let ret = new APIError('Success', 0, 200)
     try {
       const data = await REQUEST[action](...args)
       commit(COMMITS[action], data)
     } catch (e) {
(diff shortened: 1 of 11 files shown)


src/store/utils.js
```diff
@@ -1,16 +1,21 @@
...
 export function doRequestFactory(REQUEST, COMMITS) {
   return async function ({ commit }, { action, args, load = 'TOGGLE_LOADING' }) {
     if (load) commit(load, true)
+    let ret = new APIError('Success', 0, 200)
     try {
       const data = await REQUEST[action](...args)
       commit(COMMITS[action], data)
     } catch (e) {
       commit('API_ERROR', e)
+      ret = e
```
DarkPurple141: hmmmm

NunoDasNeves: The idea behind this is that the view/component/whatever can just inspect the return value now. 
Previously, you had to do the dispatch, then check the error value in the store, which would obviously break or be harder to reason about in the presence of multiple dispatches.

